PR #1, opened 2023-01-18
Refactor/move files

schouhy:
This PR renames `algebra` to `math` and moves there all the files in the `crypto` directory


MauroToscano (approved):
Great!

PR #10, opened 2023-01-27
Add BigInt and BLS12 381

ajgara:
- Adds support for generic BigInts
- In particular adds support for https://docs.rs/crypto-bigint/latest/crypto_bigint/ 
- Add BLS12 381 as a use case for bigints


math/src/elliptic_curve/element.rs
```diff
@@ -0,0 +1,218 @@
...
+use std::fmt::Debug;
+use std::marker::PhantomData;
+
+/// Represents an elliptic curve point using the projective short Weierstrass form:
+/// y^2 * z = x^3 + a * x * z^2 + b * z^3,
+/// where `x`, `y` and `z` variables are field elements.
+#[derive(Debug, Clone)]
+pub struct EllipticCurveElement<E: IsEllipticCurve> {
+    pub value: [FieldElement<E::BaseField>; 3],
+    elliptic_curve: PhantomData<E>,
```
jrchatruc, line 13: Is this `PhantomData` necessary here?

ajgara: It is not! Thank you very much.

math/src/field/fields/u384_prime_field.rs
```diff
@@ -0,0 +1,208 @@
+use std::fmt::Debug;
+use std::marker::PhantomData;
+
+use crate::field::element::FieldElement;
+use crate::field::traits::IsField;
+use crate::unsigned_integer::UnsignedInteger384 as U384;
+
+pub trait HasU384Constant: Debug + Clone + Eq + PartialEq {
+    const VALUE: U384;
+}
```
ilitteri, line 10: Let's add a comment explaining that this trait is necessary while there is no way in Rust of using a non-native const value as a generic so it could be revisited later (this will be possible in the near future).

Also, we will need a similar trait for `U256` so it would be better if we generalize this trait for the type of the order. It could be something like this: 
```suggestion
/// This trait is necessary for us to be able to use unsigned integer types bigger than 
/// `u128` (the biggest native `unit`) as constant generics.
/// This trait should be removed when Rust supports this feature.
pub trait BigOrder<T>: Debug + Clone + Eq + PartialEq {
    const VALUE: T;
}
```

This way we could use this trait for any non-native type as a constant generic for the order.

math/src/field/fields/u64_prime_field.rs
```diff
@@ -1,13 +1,13 @@
...
+use crate::cyclic_group::{HasPairing, IsCyclicGroup};
 use crate::field::element::FieldElement;
-use crate::field::traits::HasFieldOperations;
+use crate::field::traits::IsField;
 
 /// Type representing prime fields over unsigned 64-bit integers.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct U64PrimeField<const MODULO: u64>;
 pub type U64FieldElement<const ORDER: u64> = FieldElement<U64PrimeField<ORDER>>;
```
ilitteri, line 8: Is there a reason why we can't do this so we can have only one word for the same concept?
```suggestion
pub struct U64PrimeField<const ORDER: u64>;
pub type U64FieldElement<const ORDER: u64> = FieldElement<U64PrimeField<ORDER>>;
```

math/src/field/fields/u384_prime_field.rs
```diff
@@ -0,0 +1,208 @@
...
+
+/// Type representing prime fields over unsigned 64-bit integers.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct U384PrimeField<MODULO: HasU384Constant> {
+    phantom: PhantomData<MODULO>,
+}
+
+pub type U384FieldElement<ORDER> = FieldElement<U384PrimeField<ORDER>>;
+
+impl<MODULO: HasU384Constant> IsField for U384PrimeField<MODULO> {
```
ilitteri, line 20: Sames as [this comment](https://github.com/lambdaclass/lambdaworks/pull/10/files#r1100391011) here.

math/src/polynomial.rs
```diff
@@ -1,50 +1,48 @@
...
+    pub fn new(coefficients: Vec<FieldElement<F>>) -> Self {
         // Removes trailing zero coefficients at the end
         let mut unpadded_coefficients = coefficients
             .into_iter()
             .rev()
-            .skip_while(|x| *x == U64FieldElement::new(0))
-            .collect::<Vec<U64FieldElement<ORDER>>>();
+            .skip_while(|x| *x == FieldElement::zero())
+            .collect::<Vec<FieldElement<F>>>();
         unpadded_coefficients.reverse();
```
ilitteri, line 21: We could use [`take_while`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.take_while) (`*x != FieldElement::zero()`) and remove the `.rev()` step and the `unpadded_coefficients.reverse();`.

math/src/elliptic_curve/element.rs
```diff
@@ -1,19 +1,17 @@
...
-    elliptic_curve: PhantomData<E>,
+pub struct EllipticCurveElement<E: IsEllipticCurve> {
+    pub value: [FieldElement<E::BaseField>; 3],
 }
 
-impl<E: HasEllipticCurveOperations> EllipticCurveElement<E> {
+impl<E: IsEllipticCurve> EllipticCurveElement<E> {
     /// Creates an elliptic curve point giving the projective [x: y: z] coordinates.
     pub fn new(value: [FieldElement<E::BaseField>; 3]) -> Self {
         assert_eq!(
```
ilitteri, line 17: This function should return a `Result`instead of using an `assert` because doing the latter would panic in runtime. 

In general, if you don't have any choice but to panic (this is not the case tho) it is better to describe the cases in which that function would panic and why in the function documentation.

ilitteri (changes requested):
Nice work! Left some comments.

math/src/field/fields/u384_prime_field.rs
schouhy: Thanks. Added!

math/src/field/fields/u64_prime_field.rs
schouhy: No reason. Changed it.

math/src/polynomial.rs
schouhy: I think that would break for polynomials that have zeros in the middle of their representation.

ilitteri: Yes, you're right! Let's keep it this way.

PR #11, opened 2023-01-29
Implement Pippenger

jsign:
This PR includes the implementation of the Pippenger algorithm for Multi-scalar Multiplication [as described in this paper](https://eprint.iacr.org/2022/1321.pdf)).

Before this change, we only had a naive MSM implementation. I created a new `msm` module that contains:
- A `naive` module (for the original implementation) 
- A new`pippenger` module with this new implementation.

Regarding tests, I moved the existing MSM tests to the base `msm` module since all those tests should pass in all existing implementations of MSM algorithms. To do this without duplicating code or using macros, I leverage the `test-case` crate, which allows for doing things elegantly. If other MSM algorithms are added, we can add an extra `#[test_case(...)]`.

The Pippenger algorithm needs to do some SHR, SHL, and AND bitwise operations since it partitions the scalars in windows of `usize` size. I added a new trait, `IsLinearField`, which extends `HasFieldOperations`. Those bitwise operations can't be part of `HasFieldOperations` since fields like `QuadraticExtensionField` can't apply those operations. However, fields like `. In contrast, `U64PrimeField<MODULO>` _is_ a linear field.

I'll dive into some details in the PR comments.


.github/workflows/tests.yaml
```diff
@@ -0,0 +1,69 @@
+on: [push, pull_request]
```
jsign, line 1: Based on some standard flow from [actions-rs](https://github.com/actions-rs) with some tunning.

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,69 @@
...
+        uses: actions-rs/cargo@v1
+        with:
+          command: fmt
+          args: --all -- --check
+
+      - name: Run cargo clippy
+        uses: actions-rs/cargo@v1
+        with:
+          command: clippy
+          args: --all-targets -- -D warnings
```
jsign, line 69: Marking `--all-targets` since the base TOML is a workspace.

math/src/elliptic_curve/traits.rs
```diff
@@ -155,19 +155,19 @@ pub trait HasEllipticCurveOperations: Clone + Debug {
...
         } else {
             let numerator = FieldElement::from(3) * &px.pow(2) + Self::a();
             let denominator = FieldElement::from(2) * py;
             if denominator == FieldElement::zero() {
-                return qx - px;
+                qx - px
             } else {
                 let l = numerator / denominator;
-                return qy - py - l * (qx - px);
+                qy - py - l * (qx - px)
```
jsign, line 170: Remove unneeded `returns` as suggested by _clippy_.

.github/workflows/tests.yaml
jsign: I think `-D warnings` is a good target, but let me know if that sounds somewhat aggressive.

math/src/field/element.rs
```diff
@@ -92,7 +94,81 @@ where
     }
 }
 
-/// AddAssign operator overloading for field elements
+impl<F> Shl<usize> for FieldElement<F>
+where
+    F: IsLinearField,
```
jsign, line 99: We implement operations `<<`, `>>`, and `&` for `FieldElement` only if `F` is a `IsLinearField`.
(You'll see later that `IsLinearField : HasFieldOperations`).

math/src/field/fields/u64_prime_field.rs
```diff
@@ -56,6 +56,24 @@ impl<const MODULO: u64> HasFieldOperations for U64PrimeField<MODULO> {
     }
 }
 
+impl<const MODULO: u64> IsLinearField for U64PrimeField<MODULO> {
```
jsign, line 59: We implement for `U64PrimeField<MODULO>` the methods from `IsLinearField` trait.

math/src/field/fields/u64_prime_field.rs
```diff
@@ -56,6 +56,24 @@ impl<const MODULO: u64> HasFieldOperations for U64PrimeField<MODULO> {
...
+    fn bit_size() -> usize {
+        (Self::BaseType::BITS - (MODULO - 1).leading_zeros()) as usize
```
jsign, line 61: The bit size is the minimum amount of bits to represent a number in the field.
The idea is simple: "total numbers of base type" minus the "total amount of leading zeros of the biggest element".

Honestly, I'd like this method to be a `const` method since this can be resolved at compile time, but we can't have `const` methods in traits. I'm open to ideas if it sound worth it.

math/src/field/fields/u64_prime_field.rs
```diff
@@ -56,6 +56,24 @@ impl<const MODULO: u64> HasFieldOperations for U64PrimeField<MODULO> {
...
+        a >> n
+    }
+
+    fn shl(a: &Self::BaseType, n: usize) -> Self::BaseType {
+        a << n
+    }
+
+    fn and(a: &Self::BaseType, mask: usize) -> usize {
+        (*a as usize) & mask
+    }
```
jsign, line 74: It was clarifying a bit since it might not be obvious.
Since the `mask` is `usize`, the result will always fit in `usize`. (ergo, the return type is `usize`).

With the `as` conversion, we [truncate to the lower bits](https://doc.rust-lang.org/reference/expressions/operator-expr.html#semantics) and then apply the mask.

math/src/field/fields/u64_prime_field.rs
```diff
@@ -86,15 +104,10 @@ impl<const ORDER: u64> IsCyclicBilinearGroup for U64FieldElement<ORDER> {
...
     use super::*;
+    use test_case::test_case;
     const ORDER: u64 = 13;
     type FE = FieldElement<U64PrimeField<ORDER>>;
 
-    #[test]
-    fn order_must_small_as_to_not_allow_overflows() {
-        // ORDER*ORDER < u128::MAX
-        assert!(ORDER <= u64::MAX.into());
-    }
```
jsign, line 96: This had a warning from clippy that said that it made no sense. Mostly because `ORDER` is a `u64` so clearly `<= u64::MAX`. Maybe the original intention of this test was another one?

math/src/field/fields/u64_prime_field.rs
```diff
@@ -231,4 +244,48 @@ mod tests {
...
+    }
+
+    #[test]
+    fn bit_size() {
+        assert_eq!(FieldElement::<U64PrimeField<2>>::bit_size(), 1);
+        assert_eq!(FieldElement::<U64PrimeField<3>>::bit_size(), 2);
+        assert_eq!(FieldElement::<U64PrimeField<4>>::bit_size(), 2);
+        assert_eq!(FieldElement::<U64PrimeField<5>>::bit_size(), 3);
+        assert_eq!(FieldElement::<U64PrimeField<42>>::bit_size(), 6);
+    }
```
jsign, line 290: New tests for the new `IsLinearField` methods, leveraging `test-case`.

math/src/msm.rs
```diff
@@ -1,157 +0,0 @@
-use crate::cyclic_group::IsCyclicBilinearGroup;
```
jsign, line 1: Removed this file since we organize differently now in a separate `mod msm` for both this implementation, and the new one.

math/src/msm/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+pub mod naive;
+pub mod pippenger;
+
+pub trait MSM<F, G>
+where
+    G: IsCyclicBilinearGroup,
+    F: IsLinearField,
+{
+    fn msm(&self, ks: &[FieldElement<F>], ps: &[G]) -> G;
+}
```
jsign, line 14: In this new `msm` module we define the `MSM<F,G>` trait, and put all the tests that all implementations should pass.

The tests are the same that were previously used in the previous (now named `naive`) MSM algorithm, plus using `test-case` to target to both implementations (`naive` and `pipppenger`).

math/src/msm/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+            [-x, y * t, z.clone()]
+        }
+    }
+
+    type FE = FieldElement<U64PrimeField<ORDER_P>>;
+
+    #[test_case(pippenger::Pippenger::new(1))]
+    #[test_case(pippenger::Pippenger::new(2))]
+    #[test_case(pippenger::Pippenger::new(4))]
+    #[test_case(naive::Naive)]
```
jsign, line 86: Here you see how each test is targeting both implementations:
- `naive`: the previous existing one.
- `pippenger` the new one, but with different flavors of the `window_size`.

math/src/msm/naive.rs
```diff
@@ -0,0 +1,37 @@
+use crate::{
```
jsign, line 1: This file contains the same MSM implementation we had before, just moved to this new module.

math/src/msm/pippenger.rs
```diff
@@ -0,0 +1,68 @@
+use crate::{
```
jsign, line 1: This is the new Pippenger implementation.

math/src/msm/pippenger.rs
```diff
@@ -0,0 +1,68 @@
...
+
+/// Calculates the multiscalar multiplication using the Pippenger algorithm
+/// as described in section 2.3 of the paper "cuZK: Accelerating Zero-Knowledge
+/// Proof with A Faster Parallel Multi-Scalar Multiplication Algorithm on GPUs".
+/// Reference: https://eprint.iacr.org/2022/1321.pdf
+impl Pippenger {
+    pub fn new(window_size: usize) -> Self {
+        assert!(window_size > 0, "window size should be greater than zero");
+        Self { window_size }
+    }
```
jsign, line 20: A new instance of `Pippenger` is created by defining the desired `window_size`.

math/src/msm/pippenger.rs
```diff
@@ -0,0 +1,68 @@
...
+                    .scan(G::neutral_element(), |m, b| {
+                        *m = m.operate_with(b); // Reduction step.
+                        *b = G::neutral_element(); // Cleanup bucket slot to reuse in the next window
+                        Some(m.clone())
+                    })
+                    .reduce(|g, m| g.operate_with(&m))
+                    .unwrap_or_else(G::neutral_element)
+            })
+            .reduce(|t, g| t.operate_with_self(1 << self.window_size).operate_with(&g))
+            .unwrap_or_else(G::neutral_element)
```
jsign, line 66: The implementation is based is basically what's described in https://eprint.iacr.org/2022/1321.pdf.

Initially, I made a more "iterative" style implementation but ended up transforming it to a more "functional style". 
I added some extra inline comments to help bit a future reader know which part of the algorithm we're in at that point.

math/src/msm/pippenger.rs
```diff
@@ -0,0 +1,68 @@
...
+            ks.len(),
+            ps.len(),
+            "the scalar and elliptic curve point list don't have the same length"
+        );
+
+        // The number of windows of size `s` is ceil(lambda/s)
+        let num_windows = (F::bit_size() + self.window_size - 1) / self.window_size;
+
+        // We define `buckets` outside of the loop so we only have to allocate once, and reuse it
+        let mut buckets = vec![G::neutral_element(); (1 << self.window_size) - 1];
```
jsign, line 39: I defined the `buckets` to live in the heap. To avoid more than one allocation, I re-use the buckets after each "window". It as bit uglier than defining `buckets` inside the loop, but better for performance.

Honestly, I'd like to make `buckets` be an attribute in `Pippenger`, and be an array asking for a `const window_size` so it would live in the stack (i.e: no allocations). But I'm not sure what is a reasonable `window_size` and if that can put some pressure on a stack overflow. 

In any case, it's an easy change if we think is a good idea.

math/src/msm/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+
+    impl HasDistortionMap for CurrentCurve {
+        fn distorsion_map(
+            p: &[FieldElement<Self::BaseField>; 3],
+        ) -> [FieldElement<Self::BaseField>; 3] {
+            let (x, y, z) = (&p[0], &p[1], &p[2]);
+            let t = FieldElement::new([FieldElement::zero(), FieldElement::one()]);
+            [-x, y * t, z.clone()]
+        }
+    }
```
jsign, line 79: I've seen a lot of this code be moved somewhere in some draft PR (which sounds good); for now leaving it where it was.

MauroToscano:
Benchmarks with Criterion need to be added to see if the algorithm improves times

math/benches/msm.rs
```diff
@@ -0,0 +1,124 @@
...
+    },
+    msm::{pippenger, MSM},
+};
+use rand::rngs::StdRng;
+use rand::{Rng, SeedableRng};
+
+type FE = FieldElement<U64PrimeField<ORDER_P>>;
+
+macro_rules! bench_msm_for_length {
+    ( $func:ident, $a:expr ) => {
```
jsign, line 23: This is a macro that generates a criterion benchmark for a specified MSM length.
The idea is that it generates a set of N random field and group elements, and:
- Benchmarks the MSM with the Naive implementation.
- Benchmarks the MSM with Pippenger with four window-sizes (1, 2, 3 and 4).

It groups the benchmarks so they are shown nicely in the report.

math/benches/msm.rs
```diff
@@ -0,0 +1,124 @@
...
+        pub fn $func(c: &mut Criterion) {
+            let msm_size = $a;
+            let window_sizes = vec![1, 2, 3, 4];
+
+            let mut rng = StdRng::seed_from_u64(42);
```
jsign, line 28: Making the benchmark reproducible.

math/benches/msm.rs
```diff
@@ -0,0 +1,124 @@
...
+use rand::rngs::StdRng;
+use rand::{Rng, SeedableRng};
+
+type FE = FieldElement<U64PrimeField<ORDER_P>>;
+
+macro_rules! bench_msm_for_length {
+    ( $func:ident, $a:expr ) => {
+        pub fn $func(c: &mut Criterion) {
+            let msm_size = $a;
+            let window_sizes = vec![1, 2, 3, 4];
```
jsign, line 26: We can tune this to include more cases, but with this ones we get interesting insights already.

math/benches/msm.rs
```diff
@@ -0,0 +1,124 @@
...
+
+            let mut rng = StdRng::seed_from_u64(42);
+            let g = EllipticCurveElement::<CurrentCurve>::generator();
+
+            // Generate Field Elements and Group elements.
+            let s: Vec<FE> = (0..msm_size).map(|_| FE::new(rng.gen())).collect();
+            let hiding: Vec<EllipticCurveElement<CurrentCurve>> = (0..msm_size)
```
jsign, line 33: I'm using the same field element and test curves as we have today, but in the future we can expand the macro also to receive which fields and groups we want to benchmark (e.g: bls12-381, etc) and make a wider set of benchmarks in those dimensions.

math/benches/msm.rs
```diff
@@ -0,0 +1,124 @@
...
+
+            group.finish();
+        }
+    };
+}
+
+bench_msm_for_length!(msm_length_1, 1);
+bench_msm_for_length!(msm_length_10, 10);
+bench_msm_for_length!(msm_length_100, 100);
+bench_msm_for_length!(msm_length_1000, 1000);
```
jsign, line 70: We generate the benchmarks for MSM sizes 1, 10, 100 and 1000.

math/benches/msm.rs
```diff
@@ -0,0 +1,124 @@
...
+
+impl HasDistortionMap for CurrentCurve {
+    fn distorsion_map(
+        p: &[FieldElement<Self::BaseField>; 3],
+    ) -> [FieldElement<Self::BaseField>; 3] {
+        let (x, y, z) = (&p[0], &p[1], &p[2]);
+        let t = FieldElement::new([FieldElement::zero(), FieldElement::one()]);
+        [-x, y * t, z.clone()]
+    }
+}
```
jsign, line 124: I think later we can reference these test curves from other module.
Or probably just use real fields and curves.

jsign:
> Benchmarks with Criterion need to be added to see if the algorithm improves times

@MauroToscano, I included [criterion benchmarks](https://github.com/lambdaclass/lambdaworks/pull/11/commits/978eab9eb7d00ecdec0eff4d4b321d0a2b38876e) that compares both MSM in multiple dimensions.

The first one is the problem for different sizes, and the second is, for a particular MSM problem size, we compare running times for the Naive implementation and Pippenger with four different window sizes. 

In short, the setup is:
- MSM sizes: 1, 10, 100, and 1000.
- For each MSM size, compare the Naive running time with Pippenger with window sizes 1, 2, 3, and 4.

Here I leave the report results generated by criterion for all this:
image
The `msm_X` means that we're benchmarking MSM for length `X`. (e.g: if `X=3`, then that means `f1*G1+f2*G2+f3*G3`, etc).

For `X=1` (i.e: a single scalar multiplication):
image
Clearly, `Naive` is faster than Pippenger.

For `X=10`:
image

For `X=100`:
image


For `X=1000`:
image

It's quite interesting to see how the window-size has some optimal value that can be explored with these benchmarks.

Some final notes about all this:
- This is using a test field and curve. It will be useful to later switch those to real curves to make all this more meaningful.
- We should remember that the current Pippenger implementation is single-threaded, so there's room for improvement. (i.e: `rayon` or probably GPU as planned).

math/benches/msm.rs
```diff
@@ -0,0 +1,124 @@
...
+    field::{
+        element::FieldElement, fields::u64_prime_field::U64PrimeField,
+        quadratic_extension::QuadraticExtensionField,
+    },
+    msm::{pippenger, MSM},
+};
+use rand::rngs::StdRng;
+use rand::{Rng, SeedableRng};
+
+type FE = FieldElement<U64PrimeField<ORDER_P>>;
```
ilitteri, line 20: It would be better if we could change this to something more descriptive like `FieldElement`
```suggestion
type FieldElement = FieldElement<U64PrimeField<ORDER_P>>;
```

math/benches/msm.rs
```diff
@@ -0,0 +1,124 @@
...
+
+macro_rules! bench_msm_for_length {
```
ilitteri, line 22: Using a macro here isn't worth it. A more readable way of doing this instead would be defining a well-named function that parametrizes the `msm_size`.

ilitteri: Also, you should check this [benchmark setup](https://github.com/lambdaclass/AES_zero_knowledge_proof_circuit/tree/master/benches), it may be useful to see different approaches in how Criterion benchmarks could be structured.

ilitteri: And lastly, all the comments that you made in the PR about this code are really good and could be very useful as comments in the code.

MauroToscano (changes requested):
Added requests for changes

math/src/msm/pippenger.rs
```diff
@@ -0,0 +1,68 @@
...
+    window_size: usize,
+}
+
+/// Calculates the multiscalar multiplication using the Pippenger algorithm
+/// as described in section 2.3 of the paper "cuZK: Accelerating Zero-Knowledge
+/// Proof with A Faster Parallel Multi-Scalar Multiplication Algorithm on GPUs".
+/// Reference: https://eprint.iacr.org/2022/1321.pdf
+impl Pippenger {
+    pub fn new(window_size: usize) -> Self {
+        assert!(window_size > 0, "window size should be greater than zero");
```
ilitteri, line 18: This function should return a `Result`instead of using an `assert` because doing the latter would panic in runtime. 

If you don't have any choice but to panic (this is not the case tho) it is better to describe the cases in which that function would panic and why.

math/src/msm/pippenger.rs
ilitteri: This comment is good enough to be in the function's documentation for two purposes:
1. For others to understand the motivation for the implementation.
2. Nobody will ever get this information if is not on the code.

ilitteri (changes requested):
Left some comments! Ask anything if you need it.

math/src/msm/pippenger.rs
jsign: @ilitteri, now added as a comment.

math/benches/msm.rs
jsign: @ilitteri, thanks for pointing that out. This was my first time trying the reporting feature from Criterion, and I think I got confused about requiring different function names for the grouping, but now I noticed that it wasn't needed.

I removed the macros and adopted a similar style as the link you suggested. It looks like a nice setup to include benchmarks for other modules.

Also, I added some extra comments in the benchmark code.

LMK if this looks good to you.

math/Cargo.toml
```diff
@@ -7,3 +7,6 @@ edition = "2021"
 
 [dependencies]
 rand = "0.8.5"
+
+[dev-dependencies]
+test-case = "2.2.2"
```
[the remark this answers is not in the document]
jsign: Done.

.github/workflows/tests.yaml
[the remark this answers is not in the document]
jsign: @MauroToscano, sounds good, I removed it from this PR and created https://github.com/lambdaclass/lambdaworks/pull/13.

math/Cargo.toml
```diff
@@ -7,3 +7,11 @@ edition = "2021"
...
+criterion = "0.3"
+
+[[bench]]
+name = "msm"
+harness = false
```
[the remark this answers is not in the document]
jsign: Done.

math/src/msm/pippenger.rs
jsign: @ilitteri, I changed to return a `Result` now.

math/src/field/traits.rs
```diff
@@ -50,11 +50,21 @@ pub trait HasFieldOperations: Debug {
...
+
+pub trait IsLinearField: HasFieldOperations {
+    fn shr(a: &Self::BaseType, n: usize) -> Self::BaseType;
+
+    fn shl(a: &Self::BaseType, n: usize) -> Self::BaseType;
+
+    fn and(a: &Self::BaseType, mask: usize) -> usize;
+
+    fn bit_size() -> usize;
+}
```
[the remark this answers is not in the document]
jsign: > Having bit operations is not a property of `LinearFields`, so this may be confusing. And the dependency with `HasFieldOperations` is not necessary. Just implementing the std ops and asking for them in the function is enough. If you want to aggregate them in a type is good too.
> 
> The only thing missing would be `bit_size`. We may need a trait for this, or add it in another place. In any scenario, it would be better to rename it to something like `bits` or `num_bits`
> 
> The mask of the and the function should be a `BaseType` too, since the number may be bigger than an usize.

@MauroToscano, got into a rabbit hole here. 
First, some quick answers to your points:
- I changed `LinearFields` to not extend `HasFieldOperations`.
- For shift operations, I removed them from `LinearFields` and asked for those bounds in the `impl` blocks. Looks better.
- Renamed `bit_size` to `num_bits`, and also renamed `LinearFields` to `SizedField` (maybe that's a better name?).

But, I have one last thing in `SizedField` (previously called `LinearField`) which I'd want to remove: the `and` operation.

As you suggested, it would be nice to:
- Make this implementable with `std::ops` as I did with the shift operations, which worked nicely.
- Maybe make the `mask` be `Self::BaseType`, etc.

I tried to do this for some time, but I think I'm stuck trying to achieve that. The current state of this PR still works, but it needs this last extra work that I can't seem be able to do. I have a [separate branch (single commit)](https://github.com/jsign/lambdaworks/commit/4cbb49184da66779b1e5e570e789383c8c3b6c68) with the furthest I could go implementing this, but I think is a dead end.

First, I realized that trying to make `mask` be `Self::BaseType` as you suggest it shouldn't be necessary, and it complicates things a bit:
- "Not necessary" (IMHO) since `window_sizes` in Pippenger are a couple of bits long, and AFAIK big windows sizes aren't used (but I could be wrong). 
- "Complicate things a bit", since we need the result of the `&` to be `usize` since the result of that operation is used in `buckets : Vec<...>` as `buckets[and_result]` so it should be `usize`.

As mentioned before, I tried hard to be able to accommodate these constraints, but I've honestly failed to do it, probably because of my lacking Rust-type system skills!

If you can, try to pull [separate branch (single commit)](https://github.com/jsign/lambdaworks/commit/4cbb49184da66779b1e5e570e789383c8c3b6c68) (is a separate branch), and see that while I got everything to compile (i.e: `cargo build`), while doing `cargo test` I get to a fundamental problem of Rust `u64` not implementing `BitAnd<F>` for `F` that isn't strictly `u64` (i.e: `usize`).

Maybe I'm running in circles here and you can imagine an easy solution, but I think after some time of trying I'm stuck. 
Can I get a bit of help here? I'd love to see how this can be done, since this last `and` in `SizedField` is the last remaining nit to make this PR ready!

jsign:
@MauroToscano, @ilitteri, thanks for the review!

@ilitteri, all your comments were addressed so I think we have all that covered. I added some extra replies to your comments.

@MauroToscano, I got a bit into a rabbit hole regarding the `IsLinearField` removal. I got almost there, but I expanded on some problems I explained in reply to your comment. The rest of the PR comments were addressed.

math/benches/msm.rs
ilitteri: Looks good, thanks!

jsign:
@MauroToscano, this PR has been ready for re-review for some days, but I think the priority was the planning/roadmap. I expect a lot of activity soon, so this branch can get quite hairy to rebase while other more elementary constructs (more fields and curves) are getting into place.

Would you prefer I close this PR and reconsider including it again in the future?  I can't fully promise to do that, but in the worst case can serve as a reference for others. It will probably need some refactors, but big chunks of other things, like benchmarks, can be a good reference for the future. 

A bit unfortunate, but maybe we were getting ahead too fast before settling on other things.
LMK what do you think. I don't want this branch to get in the way of going full steam on other fronts.

unbalancedparentheses:
@jsign give us a few days before we answer you. we want to finish the stark prover and then we can check this and merge it.

MauroToscano:
We ended up implementing Pippenger for a newer version of the library at some point. Thank you Jsign, this was really useful.

jsign:
> We ended up implementing Pippenger for a newer version of the library at some point. Thank you Jsign, this was really useful.

No problem. Keep up with the good work.

PR #13, opened 2023-02-05
ci: add cargo workflow

jsign:
This PR is adding a CI workflow for tests and lints.

(Context: this is pulled work done initially in https://github.com/lambdaclass/lambdaworks/pull/11 but moved to a separate PR [as suggested](https://github.com/lambdaclass/lambdaworks/pull/11#discussion_r1094375590)).

cc @MauroToscano


.github/workflows/tests.yaml
```diff
@@ -0,0 +1,69 @@
+on: [push, pull_request]
+
+name: check & tests
+
+jobs:
+  check:
```
ilitteri, line 6: This job is redundant because the `cargo check` is kind of embedded in the `cargo test` command which is run in the `test` job below.

ilitteri: It would be useful to have a `fmt` job that checks the code formatting. You can use `cargo fmt -- --check` for that.

ilitteri: You could use [this](https://github.com/lambdaclass/AES_zero_knowledge_proof_circuit/blob/master/.github/workflows/ci.yml) as an example.

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,69 @@
...
+    name: Check
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
```
ilitteri, line 11: Is there a reason for not using `actions/checkout@v3`?

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,69 @@
...
+          profile: minimal
+          toolchain: stable
+          override: true
+
+      - name: Run cargo test
+        uses: actions-rs/cargo@v1
+        with:
+          command: test
+
+  lints:
```
ilitteri, line 44: It would be better to separate this into two different jobs like `linter` and `formatter` or `clippy` and `fmt`.

ilitteri (changes requested):
This PR looks good, just pointed out some minor details.

.github/workflows/tests.yaml
jsign: `cargo check` is a faster way to verify that doing `cargo build` or `cargo test`. So the intention is to have a failing job as soon as possible. 

From docs:
> This will essentially compile the packages without performing the final step of code generation, which is faster than running cargo build

If this sounds like not needed, I'm fine with removing it; but just want to make that point. Let me know if you still prefer to remove this step!

.github/workflows/tests.yaml
jsign: Nope, thanks for pointing that out. I updated to `v3`.

.github/workflows/tests.yaml
jsign: I've separated them now.

.github/workflows/tests.yaml
ilitteri: I think that you're making a good point, let's keep this job!

MauroToscano (approved):

juanbono (approved):
LGTM

PR #14, opened 2023-02-06
Implement Unsigned Integer and Montgomery Multiplication from scratch

schouhy:


math/src/elliptic_curve/traits.rs
```diff
@@ -0,0 +1,269 @@
+use crate::field::element::FieldElement;
+use crate::field::traits::IsField;
+use crate::unsigned_integer::traits::IsUnsignedInteger;
+use std::fmt::Debug;
+
+/// Trait to add elliptic curves behaviour to a struct.
+/// We use the short Weierstrass form equation: `y^2 = x^3 + a * x  + b`.
+pub trait IsEllipticCurve: Clone + Debug {
```
ilitteri, line 8: I think that it could be useful to have a `coords` or `coordinates` method for this trait. What do you think?

schouhy: There are getters for the coordinates in the `EllipticCurveElement` struct. This trait is intended to be implemented for zero-sized structs to provide specific parameters of a particular elliptic curve and provides default methods such as addition, miller, etc (which could be overridden to replace them for more efficient implementations for a particular curve if needed). There's the `EllipticCurveElement` struct to represent elements themselves of a curve. In general you would implement `IsEllipticCurve` for some zero-sized struct `MyCurve` and then instantiate elements of that curve as `EllipticCurveElement<MyCurve>`.

ilitteri: Oh, my bad, originally my suggestion was intended for the `EllipticCurveElement` struct. Now I see that the `EllipticCurveElement` has getters for individual coordinates, but I was thinking about adding a method to get the coordinates in a tuple.

schouhy: ok! done

math/src/elliptic_curve/element.rs
```diff
@@ -0,0 +1,232 @@
...
+    /// Creates an elliptic curve point giving the projective [x: y: z] coordinates.
+    pub fn new(value: [FieldElement<E::BaseField>; 3]) -> Self {
+        assert_eq!(
+            E::defining_equation(&value),
+            FieldElement::zero(),
+            "Point ({:?}) does not belong to the elliptic curve.",
+            &value
+        );
+        Self { value }
+    }
```
ilitteri, line 24: This function should return a `Result`instead of using an `assert` because doing the latter would panic in runtime. 

If you don't have any choice but to panic (this is not the case tho) it is better to describe the cases in which that function would panic and why.

ilitteri (approved):

MauroToscano (approved):

PR #16, opened 2023-02-09
`hash` module setup

ilitteri:


juanbono (approved):

MauroToscano (approved):

MauroToscano (approved):
Ok

crypto/src/hash/traits.rs
```diff
@@ -0,0 +1,19 @@
+use lambdaworks_math::field;
+
+/// Interface to Collision Resistant Hashes.
+pub trait CryptoHash<F>
```
schouhy, line 4: We are following a name convention for traits to be very distinguishable from the structs that implement them. Something like `IsXXX`  or `HasXXX` for traits or even adjectives if possible. Then the struct names would be nouns. I would name this `IsCryptoHash`. What do you think?

ilitteri: Thanks for pointing that out. I'm updating the name.

schouhy (approved):
Looks good! Thanks!

PR #17, opened 2023-02-09
Update CI config

ilitteri:
## Changes

- Update `name`.
- Update `on` to only on `push`s.
- Add concurrency.


juanbono (approved):

MauroToscano (approved):

juanbono (changes requested):
Please use the clippy target to run clippy. That ensures that we use the same as in the Makefile

juanbono (approved):

PR #35, opened 2023-02-09
WIP: Starting point for Fiat-Shamir

ajgara:


crypto/src/fiat_shamir/transcript.rs
```diff
@@ -0,0 +1,70 @@
...
+
+impl Transcript {
+    #[allow(unused)]
+    fn new() -> Self {
+        Self { data: Vec::new() }
+    }
+
+    #[allow(unused)]
+    fn append(&mut self, new_data: &[u8]) {
+        self.data.append(&mut new_data.to_owned());
```
Oppen, line 15:
```suggestion
        self.data.extend_from_slice(new_data);
```

crypto/src/fiat_shamir/transcript.rs
```diff
@@ -0,0 +1,70 @@
+use sha3::{Digest, Sha3_256};
+
+struct Transcript {
+    data: Vec<u8>,
```
Oppen, line 4: Is this expected to do anything other than hash the contents? If not, why not just keep a hasher and stream into it? It would probably be much more memory and CPU effective.

ajgara: Thanks! It looks like a good idea.

Oppen (approved):

jrchatruc (approved):

schouhy (approved):

PR #38, opened 2023-02-09
Merkle Tree Impl 

### LINKED ISSUE #24: Implement `MerkleTree` for the `hash` module

FranFiuba:
Resolves #24


crypto/src/merkle_tree/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod merkle_tree;
```
ilitteri, line 1: Newline missing also here

ilitteri:
I'll convert this PR to draft as it is in progress.

crypto/src/hash/traits.rs
```diff
@@ -9,10 +9,11 @@ where
...
     /// Hashes a field element into one. Also known as one to one hashing.
-    fn hash_one(input: field::element::FieldElement<F>) -> field::element::FieldElement<F>;
+    fn hash_one(&self, input: field::element::FieldElement<F>) -> field::element::FieldElement<F>;
 
     /// Hashes two field elements into one. Also known as two to one hashing.
     fn hash_two(
+        &self,
         left: field::element::FieldElement<F>,
         right: field::element::FieldElement<F>,
     ) -> field::element::FieldElement<F>;
```
ilitteri, line 19: The changes in this file belong to another PR. Merge `main` into this branch to fix the diff.

crypto/src/merkle_tree/merkle_tree.rs
```diff
@@ -0,0 +1,241 @@
...
+impl<F: IsField, H: IsCryptoHash<F>> MerkleTree<F, H> {
+    pub fn build(values: &[FieldElement<F>], hasher: H) -> MerkleTree<F, H> {
+        let mut leafs: Vec<TreeNode<F>> = hash_leafs(values, &hasher);
+        let mut level: Vec<TreeNode<F>> = hash_level(&mut leafs, &hasher);
+
+        while level.len() > 1 {
+            level = hash_level(&mut level, &hasher);
+        }
+
+        return MerkleTree {
```
ilitteri: This `return` is not needed.

crypto/src/hash/traits.rs
FranFiuba: fixed

crypto/src/merkle_tree/merkle_tree.rs
FranFiuba: yes fixed those kind of warnings in the following commit

ilitteri:
The PR is looking good, the `clippy` job is not passing, you can use `make clippy` to reproduce the CI's output.

crypto/src/merkle_tree/mod.rs
```diff
@@ -0,0 +1,269 @@
...
+}
+
+pub struct Proof<F: IsField, H: IsCryptoHash<F>> {
+    value: FieldElement<F>,
+    merkle_path: Vec<TreeNode<F>>,
+    hasher: H,
+}
+
+impl<F: IsField, H: IsCryptoHash<F>> Proof<F, H> {
+    pub fn verify(self, root_hash: FieldElement<F>) -> bool {
```
ilitteri, line 114: I think that it would be better to add a method in the `MerkleTree` implementation like `verify(proof: Proof<F, H>, root: FieldElement<F>) -> bool;` so then you can call it like `MerkleTree::verify`. This way the function is not related directly with the proof itself.

crypto/src/merkle_tree/mod.rs
```diff
@@ -0,0 +1,269 @@
...
+                hashed_value = self
+                    .hasher
+                    .hash_two(hashed_value, sibiling.borrow().hash.clone());
+            }
+        }
+        hashed_value == root_hash
+    }
+}
+
+type TreeNode<F> = Rc<RefCell<Node<F>>>;
```
ilitteri, line 128: Does this work if we remove `Rc` and `RefCell`?

crypto/src/merkle_tree/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+    pub fn get_proof(&self, value: FieldElement<F>) -> Option<Proof<F, H>> {
+        let hashed_value = self.hasher.hash_one(value.clone());
+
+        if let Some(leaf) = self
+            .leafs
+            .iter()
+            .find(|node| node.borrow().hash == hashed_value)
+        {
+            let mut merkle_path: Vec<TreeNode<F>> = Vec::new();
+            merkle_path = build_merkle_path(leaf.clone(), &mut merkle_path).to_vec();
```
ilitteri, line 40: These lines could be collapsed like this (if I'm not mistaken)
```suggestion
            let merkle_path = build_merkle_path(leaf.clone(), &mut Vec::new()).to_vec();
```

crypto/src/merkle_tree/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+    right.borrow_mut().sibiling = Some(left);
+    right.borrow_mut().parent = Some(parent.clone());
+
+    parent
+}
+
+fn build_merkle_path<F: IsField>(
+    node: TreeNode<F>,
+    merkle_path: &mut Vec<TreeNode<F>>,
+) -> &Vec<TreeNode<F>> {
```
ilitteri, line 110: Could we return a `Vec<TreeNode<F>>` instead of its reference? With this change you remove the need for the caller of this function to use the `to_vec()` method.

crypto/src/merkle_tree/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+
+pub struct Proof<F: IsField, H: IsCryptoHash<F>> {
+    value: FieldElement<F>,
+    merkle_path: Vec<TreeNode<F>>,
+    hasher: H,
+}
+
+type TreeNode<F> = Rc<RefCell<Node<F>>>;
+
+fn build_tree_node<F: IsField>(hash: FieldElement<F>) -> TreeNode<F> {
```
ilitteri, line 127: This function could be the method `Node::new`.

ilitteri (approved):
LGTM, nice work! We can merge after documenting.

crypto/src/merkle_tree/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+    merkle_path.to_vec()
+}
+
+pub struct Proof<F: IsField, H: IsCryptoHash<F>> {
+    value: FieldElement<F>,
+    merkle_path: Vec<TreeNode<F>>,
+    hasher: H,
+}
+
+type TreeNode<F> = Rc<RefCell<Node<F>>>;
```
pablodeymo, line 125: why do you need a RefCell?

FranFiuba: We agreed to merge this version, so that it is possible to use the merkle tree and after that I would open an other pr and refactor this implementation into a simpler version.

pablodeymo (approved):
As we talked with @FranFiuba , we can merge this PR and then refactor the code with a simpler version.

PR #44, opened 2023-02-13
Add FFT, IFFT implementation

### LINKED ISSUE #30: Implement FFT
Implement Fast Fourier transform

Example implementation: https://github.com/facebook/winterfell/blob/main/math/src/fft/mod.rs

IAvecilla:
This PR contains FFT implementation using the Cooley-Tukey algorithm. FFT is needed to transform polynomials from coefficient representation to value representation. This implementation is intended to operate over finite fields.

Closes #30.


diegokingston (approved):

ilitteri (approved):
LGTM

math/src/field/traits.rs
```diff
@@ -1,6 +1,42 @@
...
+
+/// Trait to define necessary parameters for FFT-friendly Fields.
+/// Two-Adic fields are ones whose order is of the form  $2^n k + 1$.
+/// Here $n$ is usually called the *two-adicity* of the field. The
+/// reason we care about it is that in an $n$-adic field there are $2^j$-roots
+/// of unity for every `j` between 1 and n, which is needed to do Fast Fourier.
+/// A two-adic primitive root of unity is a number w that satisfies w^(2^n) = 1
+/// and w^(j) != 1 for every j below 2^n. With this primitive root we can generate
+/// any other root of unity we need to perform FFT.
+pub trait TwoAdicField: IsField {
```
Oppen, line 14: Two things I find confusing:
1. While I do prefer the `TwoAdicField` name or those that are named after what they implement rather than a query, they already follow the "question about implementers nature" pattern, so I think we should be consistent and go with `IsTwoAdicField`.
2. Why on callers we redundantly say we want something that is both a `TwoAdicField` and an `IsField`, when the first already implies the latter?

math/src/field/test_fields/u64_test_field.rs
```diff
@@ -0,0 +1,59 @@
...
+    fn mul(a: &u64, b: &u64) -> u64 {
+        ((*a as u128 * *b as u128) % MODULUS as u128) as u64
+    }
+
+    fn div(a: &u64, b: &u64) -> u64 {
+        Self::mul(a, &Self::inv(b))
+    }
+
+    fn inv(a: &u64) -> u64 {
+        assert_ne!(*a, 0, "Cannot invert zero element");
```
Oppen, line 30: Do we have a safe version of this? Providing only panicking options is frowned upon for libraries. You should always give the caller an option to catch mistakes.

math/src/fft/helpers.rs
```diff
@@ -0,0 +1,10 @@
+use super::errors::FFTError;
+
+pub fn log2(n: usize) -> Result<u64, FFTError> {
+    if !n.is_power_of_two() {
+        return Err(FFTError::InvalidOrder(
+            "The order of polynomial should a be power of 2".to_string(),
+        ));
+    }
```
Oppen, line 8: Is rounding down not an option here?

math/src/fft/fft_cooley_tukey.rs
```diff
@@ -0,0 +1,115 @@
...
+        // Property-based test that ensures IFFT is the inverse operation of FFT.
+        #[test]
+        fn test_ifft_composed_fft_is_identity(coeffs in field_vec(8)) {
+            let poly = Polynomial::new(&coeffs[..]);
+
+            let result = fft(poly.coefficients()).unwrap();
+
+            let recovered_poly = inverse_fft(&result).unwrap();
+
+            assert_eq!(recovered_poly, poly.coefficients());
```
Oppen, line 112: Use `prop_assert_eq!`.

math/src/fft/fft_cooley_tukey.rs
```diff
@@ -0,0 +1,115 @@
...
+
+            let poly = Polynomial::new(&coeffs[..]);
+            let omega = F::get_root_of_unity(log2(poly.coefficients().len()).unwrap()).unwrap();
+
+            let result = fft(poly.coefficients()).unwrap();
+
+            let twiddles_iter = (0..poly.coefficients().len() as u64).map(|i| omega.pow(i));
+            let expected: Vec<FE> = twiddles_iter.map(|x| poly.evaluate(x)).collect();
+
+            assert_eq!(result, expected);
```
Oppen, line 98: Use `prop_assert_eq!`.

xqft: done!

math/src/fft/fft_cooley_tukey.rs
xqft: done!

math/src/field/traits.rs
jrchatruc: If I understood the second question correctly, the reason we require a `TwoAdicField` to implement `IsField` is that users need to specify a root of unity, which is an element of the field itself, so they need to provide its value using the `BaseType`

math/src/fft/helpers.rs
xqft: Could you elaborate? a hypothesis of the implemented FFT algorithm is that the amount of coefficients needs to be a power of two

math/src/field/test_fields/u64_test_field.rs
jrchatruc: Yeah we should return an error here instead, while we're at it we should do the same for `U64PrimeField` and `MontgomeryBackendPrimeField`

jrchatruc: Let's do this on a separate PR to fix them all

Jrigada (approved):

math/src/field/test_fields/u64_test_field.rs
xqft: This is now contemplated in issue #77, it can't be fixed here because it would require to change the `IsField` trait, and consequently every field implementation.

math/src/fft/helpers.rs
Oppen: Ok, so that requirement is specific to fft. I thought of log2 in general where most implementations just round down (i.e. return the index of the most significant 1 bit).

math/src/field/traits.rs
Oppen: Just for the record (since this is already merged), the second question is about the place where this trait is required, that did `TwoAdicField + IsField` for the trait bounds. Since `TwoAdicField` already has the bound `IsField`, everything after the `+` is redundant.

PR #52, opened 2023-02-13
new Polynomial reveiving slice instead of Vec

pablodeymo:
Method new should receive a slice instead a Vec allocated in the heap.

### DIFF
--- a/math/src/polynomial.rs
+++ b/math/src/polynomial.rs
@@ -16,7 +16,7 @@ impl<F: IsField> Polynomial<FieldElement<F>> {
         let mut unpadded_coefficients = coefficients
             .iter()
             .rev()
-            .skip_while(|x| **x == FieldElement::zero())
+            .skip_while(|x| x.is_zero())
             .cloned()
             .collect::<Vec<FieldElement<F>>>();
         unpadded_coefficients.reverse();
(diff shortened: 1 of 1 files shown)


math/src/polynomial.rs
```diff
@@ -11,12 +11,13 @@ pub struct Polynomial<FE> {
...
     /// Creates a new polynomial with the given coefficients
-    pub fn new(coefficients: Vec<FieldElement<F>>) -> Self {
+    pub fn new(coefficients: &[FieldElement<F>]) -> Self {
         // Removes trailing zero coefficients at the end
         let mut unpadded_coefficients = coefficients
-            .into_iter()
+            .iter()
             .rev()
-            .skip_while(|x| *x == FieldElement::zero())
+            .skip_while(|x| **x == FieldElement::zero())
```
Oppen, line 19:
```suggestion
            .skip_while(|x| x.is_zero())
```

ajgara (approved):

math/src/polynomial.rs
Oppen: The original code will create a new `FieldElement` for every iteration, while the `is_zero()` method does the check without creating new objects that could potentially allocate.

ajgara (approved):

Oppen (approved):

schouhy (approved):
Looks good! Thanks.

ajgara (approved):

Issue #37, opened 2023-02-09
Add support for Goldilocks elliptic curve


MauroToscano:
Add support for Goldilocks elliptic curve.

See [https://eprint.iacr.org/2015/625.pdf](url) for reference.

Fast Karatsuba multiplication has to be implemented for Goldilocks to be highly efficient

PR #56, opened 2023-02-13
Add the Goldilocks prime field

techiepriyansh:
Add the Goldilocks prime field with fast Karatsuba multiplication. Would be useful for implementing the Ed448-Goldilocks elliptic curve (#37).


techiepriyansh:
A couple of queries:
 * Mike Hamburg in his original C implementation suggests to unroll the loops in the Karatsuba multiplication part. Should we do the same here or is the Rust compiler smart enough to unroll the loops on its own?
 * I have only implemented the Karatsuba multiplication optimized for 64-bit machines. Do I need to implement the same for 32-bit machines too?
 * I created a new type (struct) for the internal representation of the field elements, `U56x8`. This struct currently only implements a single method for constructing it from a hex string. What additional methods does it need to suppport? In particular, does it need to conform to the `IsUnsignedInteger` trait?

ilitteri:
Could you merge from `main` so we can run the CI?

techiepriyansh:
> Could you merge from `main` so we can run the CI?
Done.

MauroToscano:
We need to add some differential fuzzing for this. Do you want to add it ? Else we can add it later in another issue.

techiepriyansh:
> We need to add some differential fuzzing for this. Do you want to add it ?

I would like to do it. But I don't really know how to do it. Can you link to me some resources on how to do it?

unbalancedparentheses:
@techiepriyansh we will it add ourselves. Give us a few days before we merge this. @MauroToscano check it and merge it once you know it's working fine

MauroToscano:
> @techiepriyansh we will it add ourselves. Give us a few days before we merge this. @MauroToscano check it and merge it once you know it's working fine

Sure

MauroToscano (approved):

schouhy (approved):
Looks good!

math/src/field/fields/p448_goldilocks_prime_field.rs
```diff
@@ -0,0 +1,304 @@
...
+        x
+    }
+}
+
+impl P448GoldilocksPrimeField {
+    fn weak_reduce(a: &mut U56x8) {
+        let a = &mut a.limbs;
+
+        let mask = (1u64 << 56) - 1;
+        let tmp = a[7] >> 56;
```
Oppen, line 162: This may be a dumb question, but if the limbs are U56, wouldn't this always be simply 0?

math/src/field/fields/p448_goldilocks_prime_field.rs
```diff
@@ -0,0 +1,304 @@
...
+        for i in (1..8).rev() {
+            a[i] = (a[i] & mask) + (a[i - 1] >> 56);
+        }
+
+        a[0] = (a[0] & mask) + tmp;
+    }
+}
+
+impl U56x8 {
+    pub const fn from(value: &str) -> Self {
```
Oppen, line 174:
```suggestion
    pub const fn from_hex_str(value: &str) -> Self {
```

Oppen (commented):
Some minor comments (one is actually a question). ~I'd also suggest adding a few proptests if possible.~

EDIT: nevermind proptest, I see differential fuzzing was discussed, which is probably a stronger proposition.

math/src/field/fields/p448_goldilocks_prime_field.rs
techiepriyansh: Sometimes, while doing certain computations, we allow the values in limbs to exceed `2^56-1` (here in the case of addition and subtraction) and then we want to reduce them again to the usual base `2^56` form (by carry propagation). This function is used for such cases and therefore `a[7] >> 56` might be non zero.

math/src/field/fields/p448_goldilocks_prime_field.rs
MauroToscano: This is the same as the rest of our code, we should fix it everywhere

Oppen: Then leave it for later :)

math/src/field/fields/p448_goldilocks_prime_field.rs
Oppen: I thought that much, but I wanted to make sure.

Oppen (approved):
All worries resolved :)

MauroToscano:
We now ask for the `FieldElement` to implement hash and a `representative`. It needs to be added before merging. But it's looking good. As soon as that is solved we can merge it

techiepriyansh:
Done. Let me know if it needs to be done differently.

MauroToscano:
Perfect. Thanks @techiepriyansh

PR #59, opened 2023-02-13
Fibonacci Stark Prover

pablodeymo:
Basic implementation of the stark proving system for the fibonacci sequence. There are a lot of `TODO`s left, which we'll tackle in other PRs, like adding fiat-shamir, zerofiers, etc.

A few things were added out of need:

- A `representative()` method for `FieldElement`s, which returns the field element's value in plain form (if the felt is in montgomery form, this returns the actual value the user expects, not its montgomery representation).
- A `compose` function to compute the composition of two polynomials.


prover/src/fri/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+}
+
+/// Returns:
+/// * new polynomoial folded with FRI protocol
+/// * new domain
+/// * evaluations of the polynomial
+pub fn next_fri_layer(
+    poly: &Polynomial<FieldElement<F>>,
+    domain: &[FE],
+    beta: &FieldElement<F>,
```
mpaulucci, line 61: I see this normally named as "alpha", did you take this nomenclature from somewhere?

prover/src/fri/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+fn fold_polynomial(
+    poly: &Polynomial<FieldElement<F>>,
+    beta: &FieldElement<F>,
+) -> Polynomial<FieldElement<F>> {
+    let coef = poly.coefficients();
+    let even_coef: Vec<FieldElement<F>> = coef
+        .iter()
+        .enumerate()
+        .filter(|(pos, _)| pos % 2 == 0)
+        .map(|(_pos, v)| *v)
```
mpaulucci, line 18: does this do anything?

prover/src/fri/mod.rs
diegokingston: Stark 101 uses beta for the randomness involved in FRI; alpha is used as the randomness to combine the polynomial constraints. Maybe we should make more comments inside the code so that it is clear.

prover/src/fri/mod.rs
pablodeymo: I am filtering the elements that are in even position, so I use enumerate que generate a tupe with (position, value), then I filter by position, and finally I need to go back to the element thrwoing away the position (with the map iterator).

prover/src/fri/mod.rs
mpaulucci: I saw beta both here https://medium.com/starkware/low-degree-testing-f7614f5172db and here https://github.com/facebook/winterfell/blob/cd76df242fcec8bab45efcdabf366fb0d8c76fbd/fri/src/folding/mod.rs#L57

diegokingston: I believe it's better to use beta in this case, so it is easier to follow from stark 101. We will make sure that there is a comment explaining each variable and we can then add a note on other naming conventions.

unbalancedparentheses: we can use any of the two but add comments and links.

jrchatruc (approved):

math/src/field/fields/u384_prime_field.rs
```diff
@@ -151,6 +151,11 @@ where
     fn from_base_type(x: Self::BaseType) -> Self::BaseType {
         MontgomeryAlgorithms::cios(&x, &C::R2, &C::MODULUS, &C::MU)
     }
+
+    // TO DO: Add tests for representatives
```
ilitteri, line 155: Let's make an issue to track this TODO

ilitteri: We could label it as "good first issue".

math/src/field/extensions/cubic.rs
```diff
@@ -127,6 +127,10 @@ where
     fn from_base_type(x: [FieldElement<Q::BaseField>; 3]) -> [FieldElement<Q::BaseField>; 3] {
         x
     }
+
+    fn representative(_x: Self::BaseType) -> Self::BaseType {
+        todo!()
```
ilitteri, line 132: [Same](https://github.com/lambdaclass/lambdaworks/pull/59/files#r1123112236) for this

math/src/field/extensions/quadratic.rs
```diff
@@ -105,6 +105,10 @@ where
     fn from_base_type(x: [FieldElement<Q::BaseField>; 2]) -> [FieldElement<Q::BaseField>; 2] {
         x
     }
+
+    fn representative(_x: Self::BaseType) -> Self::BaseType {
+        todo!()
```
ilitteri, line 110: [Same](https://github.com/lambdaclass/lambdaworks/pull/59/files#r1123112236) for this

ilitteri (approved):
LGTM

math/src/field/fields/u384_prime_field.rs
jrchatruc: Sounds good, created it here https://github.com/lambdaclass/lambdaworks/issues/138

math/src/field/extensions/cubic.rs
jrchatruc: https://github.com/lambdaclass/lambdaworks/issues/139

jrchatruc (approved):

entropidelic (approved):

entropidelic:
LGTM!

PR #64, opened 2023-02-14
Pedersen hash

### LINKED ISSUE #22: Implement `Pedersen` hash for the `hash` module

ilitteri:
Resolves #22 

This PR should be merged after #65 and #70.

TODOs:
- [x] `hash_one`.
- [x] `hash_two`.
- [ ] Test against other implementations.


crypto/src/hash/pedersen/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+pub fn random_field_element<F>(rng: &mut rand::rngs::ThreadRng) -> FieldElement<F>
+where
+    F: field::traits::IsField,
+{
+    FieldElement::<F>::from(rand::Rng::gen::<u64>(rng))
+}
+
+pub struct Pedersen<E>
+where
+    E: elliptic_curve::short_weierstrass::traits::IsShortWeierstrass,
```
igamigo, line 39: Is there any specific reason why the curve needs to be `IsShortWeierstrass` or would `IsEllipticCurve` be enough? Not sure if there is any specific requirement I'm missing.

ilitteri: Nope, I'll change it

crypto/src/hash/pedersen/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+    field::{self, element::FieldElement, traits::IsField},
+    traits::ByteConversion,
+    unsigned_integer::element::U384,
+};
+
+type FE = field::element::FieldElement<BLS12381PrimeField>;
+type Point = ProjectivePoint<BLS12381Curve>;
+
+const WINDOW_SIZE: usize = 4;
+const NUM_WINDOWS: usize = 96;
```
igamigo, line 21: Because this value is 96, `HALF_INPUT_SIZE_BITS` is 192 which means `hash_two()` is only hashing the first element (since each element is 384 and both elements are being concatenated into a single buffer).

igamigo:
I was able to instantiate the curve (and the parameter points) for the test in the branch `test-pedersen`. It also contains some smaller changes to the hash function related to my above comment.

PR #65, opened 2023-02-14
Implemented `std::iter::Sum` for `FieldElement`

### LINKED ISSUE #63: Implement `std::iter::Sum` for `FieldElement`
The task is to implement [this trait](https://doc.rust-lang.org/std/iter/trait.Sum.html) for [`FieldElement`](https://github.com/lambdaclass/lambdaworks/blob/main/math/src/field/element.rs#L8).

AleksanderMisztal:
Resolves #63


math/src/field/element.rs
```diff
@@ -103,6 +104,16 @@ where
     }
 }
 
+/// Sum operator for field elements
+impl<F> Sum<FieldElement<F>> for FieldElement<F>
+where
+    F: IsField,
+{
+    fn sum<I: Iterator<Item = Self>>(iter: I) -> Self {
+        iter.fold(Self::zero(), |a, b| a + b)
```
ilitteri, line 113: Could we change the parameters name to `addend` and `augend` like this?
```suggestion
        iter.fold(Self::zero(), |augend, addend| augend + addend)
```

ilitteri (changes requested):
The PR looks good! Just one minor change and we can merge!

ilitteri (approved):
LGTM

AleksanderMisztal:
Btw I have some spare time over the next few days, I could tackle something bigger. What's the best way to message you to discuss?

ilitteri:
> Btw I have some spare time over the next few days, I could tackle something bigger. What's the best way to message you to discuss?

First of all thanks for addressing the issue.

I think it'd be awesome if you tackle something bigger.

Regarding your question, I'll @unbalancedparentheses and @MauroToscano answer it.

jrchatruc (approved):

unbalancedparentheses:
Hi! The best way is that you join the telegram and we talk over there. You can find the telegram group in the readme!

AleksanderMisztal:
Thanks! But looks like that link is broken 👀

ilitteri:
> Thanks! But looks like that link is broken 👀

Fixed!

ilitteri:
Could you merge from `main` so we can run the CI?

ilitteri:
Sorry for the dealy, just one more detail. Could we add a test for this?

MauroToscano:
There is a small conflict, if it's fix we can merge it

MauroToscano (approved):

PR #71, opened 2023-02-14
Add elliptic curve trait

ajgara:
- Add a generic trait for all the types of curves. (ShortWeierstrass, Edwards, Montgomery, etc...).
- Refactor in PointRepresentation and EllipticCurve representation to make it easier for the users.

### DIFF
--- a/math/src/elliptic_curve/short_weierstrass/curves/bls12_377/curve.rs
+++ b/math/src/elliptic_curve/short_weierstrass/curves/bls12_377/curve.rs
@@ -27,7 +27,7 @@ impl IsEllipticCurve for BLS12377Curve {
         ])
     }
 
-    fn create_affine_point(
+    fn create_point_from_affine(
         x: FieldElement<Self::BaseField>,
         y: FieldElement<Self::BaseField>,
     ) -> Self::PointRepresentation {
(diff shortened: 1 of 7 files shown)


Oppen (approved):

math/src/elliptic_curve/short_weierstrass/curves/bls12_377/curve.rs
```diff
@@ -12,8 +14,28 @@ const BLS12377_MAIN_SUBGROUP_ORDER: U384 =
 /// The description of the curve.
 #[derive(Clone, Debug)]
 pub struct BLS12377Curve;
-impl IsShortWeierstrass for BLS12377Curve {
+
+impl IsEllipticCurve for BLS12377Curve {
     type BaseField = BLS12377PrimeField;
+    type PointRepresentation = ProjectivePoint<Self>;
+
+    fn generator() -> Self::PointRepresentation {
```
ilitteri, line 22: I think that it'd be a good idea for the `generator` function should be `const` in the future. We can file an issue for this.

math/src/elliptic_curve/short_weierstrass/curves/bls12_377/curve.rs
```diff
@@ -96,21 +110,21 @@ mod tests {
 
     #[test]
     #[should_panic]
-    fn create_invalid_points_panicks() {
-        ProjectivePoint::<BLS12377Curve>::new([FEE::from(1), FEE::from(1), FEE::from(1)]);
+    fn create_invalid_points_panics() {
```
ilitteri, line 113: We should evaluate if it still makes sense to have this test, given that it'll only panic only in debug idk if it is worth testing.

math/src/elliptic_curve/short_weierstrass/curves/bls12_377/curve.rs
ilitteri: Or at least to have a `const` version.

ilitteri (approved):
LGTM, just left some comments to discuss.

jrchatruc (approved):

schouhy:
@ilitteri We are working on another PR that includes other refactors as well. We'll add your suggestions in that PR.

PR #80, opened 2023-02-15
Add benchmark for u64 field

AminArria:


math/Cargo.toml
```diff
@@ -8,3 +8,8 @@ edition = "2021"
 [dependencies]
 rand = "0.8.5"
 thiserror = "1.0"
+criterion = "0.4"
```
AminArria, line 11: Move to `dev-dependencies`

ilitteri (approved):

MauroToscano:
There should be a README, either in the main one or in the module, that says how to run the benchmarks.

 A make command to run all benches may be useful too.

MauroToscano (changes requested):
Add documentation on how to run the benches. Add a command in the makefile to run all benches, and write in the main README how to run all benches or some of them.

MauroToscano (approved):

PR #82, opened 2023-02-15
feat: compute composite polynomial from trace using modified winterfell

entropidelic:
This PR implements functionality to get a composition and trace polynomials from a trace and an AIR, using a modified version of winterfell.
* Implemented the function `get_cp_and_tp`, that takes as input a trace and an AIR as defined by winterfell data structures, and outputs composition and trace polynomials as our `Polynomial` data structure.
* Implemented a test with the example in the [winterfell readme](https://github.com/facebook/winterfell/#usage)


prover/src/fri/mod.rs
```diff
@@ -0,0 +1,11 @@
...
+    println!("Hello world");
+}
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn test_hello() {
+        super::hello();
+    }
+}
```
MauroToscano, line 11: This code should be deleted

math/src/unsigned_integer/element.rs
```diff
@@ -9,6 +9,7 @@ use std::fmt::Debug;
 
 pub type U384 = UnsignedInteger<6>;
 pub type U256 = UnsignedInteger<4>;
+pub type U128 = UnsignedInteger<2>;
```
MauroToscano, line 12: This is unused

MauroToscano (changes requested):

PR #84, opened 2023-02-15
Implement `Arbitraty` trait in all types

### LINKED ISSUE #40: Implement Arbritrary for all types
Building from the ground up with fuzzing in mind is always beneficial to libraries, especially ones that define their own types.

Meaning the [Arbitrary trait](https://github.com/rust-fuzz/arbitrary) should be implemented for all types, maybe even on the features themselves. This way, setting up `libFuzzer`, `cargo fuzz`, or `AFL` is easy. 

The standard approach is to add a feature gate `arbitrary` that reveals the `Arbitrary` trait implementations when enabled.

ilitteri:
This PR is related and intends to resolve #40

### DIFF
--- a/crypto/Cargo.toml
+++ b/crypto/Cargo.toml
@@ -4,6 +4,10 @@ version = "0.1.0"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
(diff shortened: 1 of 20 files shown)


Oppen (approved):

MauroToscano:
Stale

PR #89, opened 2023-02-16
Make `IsCryptoHash` trait return `Result`

igamigo:
- Change hash trait to return results instead of elements


ilitteri (approved):
LGTM, are the defined errors related to Poseidon or they are just generic ones?

PR #92, opened 2023-02-17
Benchmark for FFT

AminArria:


math/benches/fft.rs
```diff
@@ -0,0 +1,108 @@
+use const_random::const_random;
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+use lambdaworks_math::fft::fft_cooley_tukey::{fft, inverse_fft};
+use lambdaworks_math::field::element::FieldElement;
+use lambdaworks_math::field::traits::{IsField, IsTwoAdicField};
+use rand::Rng;
+
+// Copied from math/src/field/test_fields/u64_test_field.rs
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct U64TestField<const MODULUS: u64>;
```
AminArria, line 10: This was needed because the one defined in `u64_test_field.rs` is not exported outside the crate, should we export it instead to use here?

MauroToscano: We can make it Pub, since its under test fields and people will not confuse it with common fields, as to avoid the duplicated code.

math/benches/fft.rs
```diff
@@ -0,0 +1,108 @@
...
+fn fft_benchmark(c: &mut Criterion) {
+    const MODULUS: u64 = PRIMES[const_random!(usize) % PRIMES.len()];
+    c.bench_function("fft", |bench| {
+        let mut rng = rand::thread_rng();
+        let coeffs_size = 1 << rng.gen_range(1..8);
+        let mut coeffs: Vec<FieldElement<U64TestField<MODULUS>>> = vec![];
+
+        for _ in 0..coeffs_size {
+            coeffs.push(FieldElement::new(rng.gen_range(1..=u64::MAX)));
+        }
```
ilitteri, line 84: Aren't these lines being benchmarked too? Or only what's inside `bench.iter`?

math/benches/fft.rs
AminArria: Made it public

jrchatruc (approved):

ilitteri (approved):
LGTM

MauroToscano (approved):

PR #102, opened 2023-02-17
Add Flamegraphs to the benchmarks

ilitteri:
### Generating a flamegraph for a specific benchmark

To generate a flamegraph for a specific benchmark, run:

```bash
make flamegraph BENCH=bench_name
```

You can check the generated flamegraph in `target/criterion/`

### Generating a flamegraph for all benchmarks

To generate a flamegraph for all benchmarks, run:

```bash
make flamegraphs
```

### Viewing the flamegraph

We will now find a file called `flamegraph.svg` in `target/criterion/<bench_name>/profile/flamegraph.svg`


jrchatruc (approved):

Oppen (approved):
LGTM in principle. However, before merging I would like to know if this way has any advantages over running `cargo flamegraph --root --bench -- --bench ${BENCH}`. That one would be transparent, with no code changes needed.

MauroToscano:
This is outdated and stale. We can review it later if we want to add flamegraphs like this

PR #101, opened 2023-02-17
Restructure benchmarks module

ilitteri:
Solves #91. This PR should be merged after #92.

## Run all the benchmarks
```
make benchmarks
```

## Run specific benchmarks
```
make benchmark BENCHMARK=benchmark_name
```

## Adding a benchmark
To add a benchmark you just need to:
1. Go to `some_crate/Cargo.toml`
2. Declare your new bench:
  ```toml
  [[bench]]
  name = "<my_new_bench>"
  harness = false
  ```
3. Create `benches/<my_new_bench>_benchmarks.rs`
4. Copy the structure of another bench.
5. Define your benchmark functions in `benches/benchmarks/my_new_bench.rs`

PR #110, opened 2023-02-20
Add benchmark for polynomial

AminArria:
# Benchmark for polynomial module

## Description

Adds benchmarks for polynomial module

Requires #101 to be merged first and a rebase on `main` afterwards


math/benches/polynomial_benchmarks.rs
```diff
@@ -0,0 +1,10 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+
+mod benchmarks;
+
+fn run_polynomial_benchmarks(c: &mut Criterion) {
+    benchmarks::polynomial::polynomial_benchmark(c);
```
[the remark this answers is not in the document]
AminArria: Added

ilitteri (approved):
LGTM

math/benches/benchmarks/polynomial.rs
```diff
@@ -0,0 +1,78 @@
...
+    });
+
+    group.bench_function("add", |bench| {
+        let x_poly = Polynomial::new(&gen_fe_vec());
+        let y_poly = Polynomial::new(&gen_fe_vec());
+        bench.iter(|| black_box(&x_poly) + black_box(&y_poly));
+    });
+
+    // The ones below can't be done because reference operations are not implemented and not having
+    // references causes E0507 because FnMut closure tries to move because Polynomial does not implement Copy
```
MauroToscano, line 42: There must be a way to benchmark this, most of structs don't implement copy. There is clone if you need to copy them, but I'm not sure what's the issue.

MauroToscano: We can do that in Another PR, but I'd remove the functions if they are not working and just comment which operations are not benchmarked

MauroToscano (changes requested):

math/benches/benchmarks/polynomial.rs
AminArria: I didn't immediately clone because I thought it could affect the results, but maybe it can be negligible and if we always do it is consistent across runs (same overhead)

With `clone()` everything works, should I use that then?

ilitteri: Let's use `clone()`, we can make flamegraphs after to see how using `clone()` impacts in the benchmark.

AminArria: Added the `clone()` and uncommented the other benchmarks

jrchatruc (approved):

IAvecilla (approved):

MauroToscano (approved):

Issue #25, opened 2023-02-09
Random Field Element


MauroToscano:
Implement the `rand` traits for `FieldElement<F>` where `F: IsFIeld`.

gluax:
UX should be considered here more. Rather than just defining a `random` method that uses a generic as an argument, we should implement the `rand` traits for the types.

It is more convenient and better for user experience when using the library.

ilitteri:
> UX should be considered here more. Rather than just defining a `random` method that uses a generic as an argument, we should implement the `rand` traits for the types.
> 
> It is more convenient and better for user experience when using the library.

Thanks for the suggestion, you're right, we'll update the description!

ilitteri:
> Can I take it?

Sure!

GianfrancoBazzani:
Hi,  i cannot find `IsFielElement` trait should it be included in `IsField`?

ilitteri:
> Hi, I cannot find `IsFielElement` trait, should it be included in `IsField`? or maybe `IsField` should be renamed to `IsFielElement`? Sorry for my ignorance, I'm not familiar with rust naming conventions yet.

You're right! `IsFieldElement` was the former name of `IsField`. You should implement it for `IsField`.

ilitteri:
> Hi, I cannot find `IsFielElement` trait, should it be included in `IsField`? or maybe `IsField` should be renamed to `IsFielElement`? Sorry for my ignorance, I'm not familiar with rust naming conventions yet.

The description is updated.

MauroToscano:
This is done, though the code is not clear on that, we can create another issue to move it around

PR #115, opened 2023-02-22
Implement rand generation for field elements

GianfrancoBazzani:
# Implement rand generation for field elements

## Description

- solves https://github.com/lambdaclass/lambdaworks/issues/25 .

- Implemented following rust documentation [link](https://rust-lang-nursery.github.io/rust-cookbook/algorithms/randomness.html#generate-random-values-of-a-custom-type) 
- For the felt u64 i implemented a test that checks if the values are generated following a normal distribution (maybe it is not necessary).

Randoms can be generated with rand::random() method or directly calling gen() method of a ThreadRng object .
examples:

```
 let rand: F23Element = rand::random();
```

or

```
 let mut rng:ThreadRng = rand::thread_rng();
 let rand: FP1Element = rng.gen();
```

## Type of change

- [x] New feature

## Checklist
- [x] Linked to Github Issue
- [x] Unit tests added


math/src/field/fields/u384_prime_field.rs
```diff
@@ -126,8 +128,51 @@ where
...
+
+        let integer: U384 = U384::from(&rand_hex);
+
+        FieldElement::new(integer)
+    }
+}
+
 #[cfg(test)]
 mod tests {
+
```
ilitteri, line 175: Let's remove this line.

GianfrancoBazzani:
I definitely think that it can be generalized following the same method as here: https://github.com/lambdaclass/lambdaworks/issues/126

math/src/field/fields/montgomery_backed_prime_fields.rs
```diff
@@ -197,6 +199,52 @@ where
...
+                    is_msl = false;
+                }
+                _ if (i == 5) => {
+                    if rand_limbs == [0; NUM_LIMBS] {
+                        //to avoid 0 in the generation set
+                        rand_limbs[i] = rng.gen_range(1..=u64::MAX);
+                    } else {
+                        rand_limbs[i] = rng.gen_range(0..=u64::MAX);
+                    }
+                }
```
MauroToscano, line 231: This 5 seems a bit off. If it's NUM_LIMBS - 1 this case should never be reached, because is_msl is true and the first case matches.

math/src/field/fields/montgomery_backed_prime_fields.rs
```diff
@@ -197,6 +199,52 @@ where
...
+    ) -> FieldElement<MontgomeryBackendPrimeField<C, NUM_LIMBS>> {
+        let mut rand_limbs: [u64; NUM_LIMBS] = [0; NUM_LIMBS];
+        let mut is_msl: bool = true; //is Most Significative Limb of MODULUS
+
+        for i in 0..NUM_LIMBS {
+            match C::MODULUS.limbs[i] {
+                mod_limb if (mod_limb == 0) && is_msl => rand_limbs[i] = 0,
+                mod_limb if (i == (NUM_LIMBS - 1)) && is_msl => {
+                    rand_limbs[i] = rng.gen_range(1..mod_limb)
+                }
```
MauroToscano, line 219: This match can't be reached, it's on the last iteration, and the first one with an is_msl will turn the flag off

MauroToscano:
The algorithm seems a bit off. I would expect to see something like ->

```
has_to_roll_highest_mod = true
for each limb starting from the most significative
   if has_to_roll_highest_mod 
     Roll up to maximium number of the limb
     if has_to_roll_highest_mod
       has_to_roll_highest =  (rolled_number == max_limb)
   else
     roll highest u64
```

math/src/field/fields/montgomery_backed_prime_fields.rs
GianfrancoBazzani: Sorry, my bad, the algo was originally designed for 6 limbs felts just fixed.

math/src/field/fields/montgomery_backed_prime_fields.rs
```diff
@@ -197,6 +199,40 @@ where
...
+        let mut rand_hex: String = "".to_string();
+
+        for i in rand_limbs {
+            rand_hex.push_str(&format!("{i:0>16X}"));
+        }
+
+        let integer: UnsignedInteger<NUM_LIMBS> = UnsignedInteger::from(&rand_hex);
+
+        FieldElement::new(integer)
+    }
```
MauroToscano, line 233: Yo should be able to create the UnsignedInt from the limbs directly. Knowing that 
"/// The most significant bit is in the left-most position.
/// That is, the array `[a_n, ..., a_0]` represents the"

Create the unsigned int with

using UnsignedInt {
  limbs: rand_limbs
}

GianfrancoBazzani: Done!

math/src/field/fields/montgomery_backed_prime_fields.rs
```diff
@@ -197,6 +199,32 @@ where
...
+        let mod_min_one = C::MODULUS - UnsignedInteger::from_u64(1_u64);
+
+        for (i, rand_limb) in rand_limbs.iter_mut().enumerate().take(NUM_LIMBS) {
+            if !roll_to_max {
+                *rand_limb = rng.gen_range(0..=mod_min_one.limbs[i]);
+                roll_to_max = *rand_limb != mod_min_one.limbs[i];
+            } else {
+                *rand_limb = rng.gen_range(0..=u64::MAX);
+            }
+        }
```
MauroToscano, line 222: One last thing, this for would be easier if it's just a for i, and we can quickly check the endianess

MauroToscano (changes requested):

jrchatruc (approved):

math/src/field/fields/u64_prime_field.rs
```diff
@@ -249,4 +260,75 @@ mod tests {
...
+            }
+
+            // Compute arithmetic average of histogram columns values
+            let mut avg: u64 = 0;
+
+            for i in hist {
+                avg += i;
+            }
+            avg /= hist.len() as u64;
+
```
pablodeymo, line 298: this could be written as
`let avg = hist.iter().sum() / hist.len() as u64;`

math/src/field/fields/u64_prime_field.rs
```diff
@@ -249,4 +260,75 @@ mod tests {
...
+        let mut sigma_probs: [f64; 4] = [0.0; 4];
+
+        for _ in 0..n {
+            // Generate a rand distribution of field elements
+            const N_SAMPLES: usize = 200000;
+            let mut rand_dist: [FEx; N_SAMPLES] = [FEx::zero(); N_SAMPLES];
+
+            for i in &mut rand_dist {
+                *i = rng.gen();
+            }
```
pablodeymo, line 282: this could be better done in the initialization of rand_dist

GianfrancoBazzani: tried the following codes but both panics with stack overflow during execution for large values of N_SAMPLES
```rust
let rand_dist: [FEx; N_SAMPLES] = [0; N_SAMPLES].map(|_| rng.gen());
```
and

```rust
let rand_dist: [FEx; N_SAMPLES] = {
     let mut z_array = [FEx::zero(); N_SAMPLES];
     for i in &mut z_array {
          *i = rng.gen();
     }
     z_array
};
            
```

math/src/field/fields/u64_prime_field.rs
GianfrancoBazzani: Done!

MauroToscano:
After an internal discussion we decided to implement a secure hash to Field algorithm. See issue #260

math/src/field/fields/montgomery_backed_prime_fields.rs
```diff
@@ -208,6 +210,33 @@ where
...
+{
+    fn sample<R: Rng + ?Sized>(
+        &self,
+        rng: &mut R,
+    ) -> FieldElement<MontgomeryBackendPrimeField<M, NUM_LIMBS>> {
+        let mut rand_limbs: [u64; NUM_LIMBS] = [0; NUM_LIMBS];
+        let mut roll_to_max: bool = false;
+        let mod_min_one = M::MODULUS - UnsignedInteger::from_u64(1_u64);
+
+        #[allow(clippy::all)] //Lint imposed by clippy makes code harder to read.
```
Oppen, line 226: `clippy::all` seems excessive. Try to limit it to the actual lint that causes problems.

math/src/field/fields/u64_prime_field.rs
Oppen: Since this is a test, you could use a `Vec<_>`. Don't need to squeeze the last drop of performance in this case. So, I'd suggest something like:
```rust
let rand_dist: Vec<FEx> = (0..N_SAMPLES).map(|_| rng.gen()).collect();
```
This should allocate on the heap, avoiding the stack overflow.

math/src/field/fields/u64_prime_field.rs
```diff
@@ -281,4 +292,73 @@ mod tests {
         let f2 = FE::new(f1.representative());
         assert_eq!(f1, f2);
     }
+
+    #[test]
+    fn generate_random_using_rng_generator_and_check_if_normal_distribution() {
```
Oppen, line 297: Aren't PRNGs expected to be uniform rather than normal?

Oppen: This is extracted from the linked documentation:

> By default, random numbers in the rand crate have [uniform distribution](https://en.wikipedia.org/wiki/Uniform_distribution_(continuous)). The [rand_distr](https://docs.rs/rand_distr/*/rand_distr/index.html) crate provides other kinds of distributions. To use them, you instantiate a distribution, then sample from that distribution using [Distribution::sample](https://docs.rs/rand/*/rand/distributions/trait.Distribution.html#tymethod.sample) with help of a random-number generator [rand::Rng](https://docs.rs/rand/*/rand/trait.Rng.html).

GianfrancoBazzani: Sorry i think the name of the tests is not completely descriptive, i try to test that the random generator is uniform by applying the [central limit theorem](https://en.wikipedia.org/wiki/Central_limit_theorem). What has a Gaussian shape is the distribution of the histogram counts for a large quantity of samples taken.  IDNK if this is a correct/good way to test it.

Oppen: But doesn't the page say the requirement is that the samples are identically distributed (as in, all samples have the same distribution) rather than the elements in the sample having uniform distribution? I think CLT applies for many distributions other than uniform, even if that is included in the set.

GianfrancoBazzani: After checking it I think I was wrong, CLT applies to the distribution of the samples mean not to the distribution of the samples. Therefore, doing that, even if the generator will generate a nonuniform distribution, the test will pass so as you say that applies  for many distributions of the samples. A correct way the correct way would have been to apply chi-square test. Thanks for that! Anyway, this PR will be closed without merge

Issue #130, opened 2023-03-01
Add boundary polynomial functionality


entropidelic:
* Interpolate trace values where there are boundary constraintsto obtain polynomial
* Subtract this new polynomial to the trace polynomial
* Divide by zerofier

PR #136, opened 2023-03-02
Add functionality to compute boundary quotient

entropidelic:
ref #130
 
This PR adds a new `constraints` module, and a `boundary` module for boundary constraints inside.
Helper function `get_boundary_quotient` is added in the proving-system module, to get the boundary quotient used in the protocol.

TODOs:
* ~~Quotient still needs to be divided by zerofier~~
* ~~Integrate in prover~~


jrchatruc (approved):

proving-system/stark/src/constraints/boundary.rs
```diff
@@ -0,0 +1,117 @@
...
+        Self { constraints }
+    }
+
+    pub(crate) fn get_steps(&self) -> Vec<usize> {
+        let mut steps = Vec::with_capacity(self.constraints.len());
+        for constraint in self.constraints.iter() {
+            steps.push(constraint.step);
+        }
+
+        steps
```
ilitteri, line 49: This could also be done with
```suggestion
        self.constraints.iter().map(|constraint| constraint.step).collect()
```

ilitteri (approved):

MauroToscano:
Can we remove all the gets ? It's a weird convention that's not used in Rust

MauroToscano (changes requested):
Functions should be documented. Struct should have a small description of what it is.

MauroToscano (approved):

Issue #126, opened 2023-02-27
Generalize `IsMontgomeryConfiguration`  and `MontgomeryBackendPrimeField`


GianfrancoBazzani:
In order to solve https://github.com/lambdaclass/lambdaworks/issues/95 without repeat code i tried to generalize `IsMontgomeryConfiguration`  trait and  `MontgomeryBackendPrimeField` struct to allow usage with any `UnsignedInteger`.

Tried something like this:
![image](https://user-images.githubusercontent.com/55500596/221582398-a973d1ec-b6a7-4197-8e90-26bcc7b45f30.png)

If  i use  a generic U: `IsUnsignedInteger` :

![image](https://user-images.githubusercontent.com/55500596/221582869-381b0f54-729b-416d-bf1e-1bbb8f33d13f.png)

Any idea to define a generic <U> to generalize any type `UnsignedInteger<_>` struct like?

MauroToscano:
Don't generalize it over the trait, we just need it to work with UnsignedInteger<u: usize> for now. This will enable the auto paramater generator we added to work smoothly, since rust don't like const functions in traits.

MauroToscano:
This backend is to make generally optimized code and enable a fast implementation of any field. The IsUnsignedInteger is for everything to keep working if you want to make a really specific optimization, or swap the native type for another one.

MauroToscano:
Leaving that aside, the problem there is that you are using a trait, so you can't use `UnsignedInteger` functions directly if you define the variable as `U`. You are limited to the trait functions, because the U could be something different.

GianfrancoBazzani:
Tanks a lot Mauro, you are truly a Rust warlock. So instead of generalize for any object with `IsUnsignedInteger` trait implementation, I generalized for any `NUM_LIMBS` constant value and used `UnsignedInteger<NUM_LIMBS>` to generalize for any object following custom multiprecision unsigned integer struct like (Types: U256, U384,....). Seems Works and passes the tests:

![Screenshot from 2023-02-27 15-55-39](https://user-images.githubusercontent.com/55500596/221598653-845496a1-4cb2-4459-9c77-9f2049828f33.png)

GianfrancoBazzani:
> Leaving that aside, the problem there is that you are using a trait, so you can't use `UnsignedInteger` functions directly if you define the variable as `U`. You are limited to the trait functions, because the U could be something different.

[...]

Issue #95, opened 2023-02-17
Implement `U256PrimeField`


ilitteri:

GianfrancoBazzani:
Can i take it?

MauroToscano:
Would be nice. 

There is an issue that the Montgomery configuration is inside the U384 PrimeField, and we should use it in the U256 PrimeField too.

To make it easier to implement this issue, the Montgomery Backend could be replicated for U256, and generalized later. Or if it's not too complicated, generalized along with the issue.

And right now the Montgomery backend needs you to input MP and R2. If you need to create test fields, you can calculate R2 and MP with the following formulas in Python:

```
MP = pow(-MODULUS, -1, 2**64)
R2 = 2**(256 * 2) % MODULUS
```

This should be improved, we are working on it.

GianfrancoBazzani:
Thanks a lot! I was wondering how to get the Montgomery conf values!

MauroToscano:
This should help @GianfrancoBazzani 

https://github.com/lambdaclass/lambdaworks/pull/118

The algorithms can be adapted for arbitrary sizes of limbs, including 256

MauroToscano:
Solved

PR #137, opened 2023-03-02
Implement u256 prime field

GianfrancoBazzani:
# Implement u256 prime field

## Description

solves https://github.com/lambdaclass/lambdaworks/issues/126
solves https://github.com/lambdaclass/lambdaworks/issues/95

Implemented U256PrimeField and generalized Montgomery backed for avoiding code repetition and allow easy and fast future implementations of other fields with different limb lengths.

Example of the UX while using the mod:
Screenshot from 2023-03-02 18-25-44

trait aliases are experimental see issue #41517 <https://github.com/rust-lang/rust/issues/41517>, but in the future the use of trait alias may upgrade the UX to something like this, note that instead of use general trait IsMontgomeryConfiguration<"limbs"> a trait alias would be defined, for example U256IsMontgomeryConfiguration =  IsMontgomeryConfiguration<4>:

Screenshot from 2023-03-02 18-28-53

## Type of change

- [x] New feature

## Checklist
- [x] Linked to Github Issue
- [x] Unit tests added


MauroToscano (approved):
Looks good!

As a side note, use git move when renaming files, so git realizes it's the same one edited.

jrchatruc (approved):

ilitteri (approved):

PR #148, opened 2023-03-03
Security vulns on Stark Verifier examples

MauroToscano:
# Collection of exploits to be fixed on the Stark Verifier

## Description

Malicious user action of exploit i:

1. Send elements not in paths 
2. Send elements not in the queried index
3. Send malicious trace

1 and 2 are Related to issues in the MerkleTree implementation and usage

PR #152, opened 2023-03-06
Remove bools from Merkle Proof struct

### LINKED ISSUE #149: MerkleTree `proof` `verify` shouldn't use saved bools to know how to hash
Proof verify is using saved bools to know if it should hash to the right or the left. It's required that the path only contains the FieldElements, and the verification deduces how to hash it from an index in the merkle tree. 

This allows the caller of the verify to check the element corresponds to the expected position

jrchatruc:
Closes #149.

Merkle Proofs no longer store booleans to know which index to hash when verifying. Instead, the `verify` method now takes an index as argument and deduces which side it should hash on.

This PR fixes the second vulnerability described in #148.

## Type of change

- [x] New feature
- [x] Bug fix

## Checklist
- [x] Linked to Github Issue


MauroToscano (approved):

entropidelic (approved):

ilitteri (approved):
LGTM

PR #155, opened 2023-03-06
Only require approvals from `@lambdaclass/zk_research_and_development`

mpaulucci:
## Description
Let's add @schouhy and @ajgara to the zk_research_and_development, otherwise we will need approvals from the group and both of them.


MauroToscano:
This can't be done because they are not part of LambdaClass

MauroToscano:
Also, you don't need approvals from them, you need three of either zk r&d or them

PR #190, opened 2023-03-17
Implemented POC of parallel FFT and twiddle generation in Metal

xqft:
Implemented POC of parallel FFT in Metal

## Description

This PR creates a new `gpu` crate and adds a proof-of-concept implementation of parallel FFT and twiddle generation in Metal. Currently it only supports `u32` prime fields. Benchmarks suggest that parallel FFT in Metal is ~14x as fast as the actual sequential implementation.

A follow-up PR will be made for developing an API around this POC and other developments for expanding and making it usable within the library.

## Type of change

- [x] New feature
- [x] Optimization

## Checklist
- [x] Linked to Github Issue
- [x] Unit tests added
  - [x] Benchmarks added/run


xqft:
Project is failing to compile I think because the CI's container doesn't have an obj-c cross-compiler

Cargo.toml
```diff
@@ -4,4 +4,5 @@ members = [
     "math",
     "crypto",
     "proving-system/stark",
+    "gpu",
```
MauroToscano, line 7: Is it going to be shared between multiple GPUs ? Else I'd call this metal. Or metal GPU. Or it can be like proving-system, where you have a folder with multiple workspaces, one for each proving system

gpu/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "lambdaworks-gpu"
```
MauroToscano, line 2: Is it going to be shared between multiple GPUs ? Else I'd call this lambdaworks-metal. Or lambdaworks-metal-gpu

math/src/field/test_fields/u64_test_field.rs
```diff
@@ -1,9 +1,9 @@
 use crate::field::traits::{IsField, IsPrimeField, IsTwoAdicField};
 
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct U64TestField<const MODULUS: u64>;
+pub struct U64Field<const MODULUS: u64>;
```
MauroToscano, line 4: Why this change ? There is an u64PrimeField for non testing cases. This part of the library should be used only for tests, not making that explicit may lead to confusion for users of the library. If you want to change the name wherever you use it, you can always use as wherever you import it

MauroToscano (changes requested):

math/src/fft/operations.rs
IAvecilla, line 1: Do you think this file is necessary? We already implemented all the FFT operations in the `Polynomial.rs` file. This might be deprecated now.

xqft: Yep it´s useless now, fixed!

gpu/src/metal/fp.h.metal
```diff
@@ -0,0 +1,214 @@
...
+}
+
+/// Compute the multiplicative inverse of x, or `1/x` in finite field terms.  Since `x^(P-1) == 1
+/// (mod P)` for any x != 0 (as a consequence of Fermat's little therorm), it follows that `x *
+/// x^(P-2) == 1 (mod P)` for x != 0.  That is, `x^(P-2)` is the multiplicative inverse of x.
+/// Computed this way, the 'inverse' of zero comes out as zero, which is convient in many cases, so
+/// we leave it.
+constexpr inline Fp inv(Fp x) {
+  return pow(x, Fp::P - 2);
+}
```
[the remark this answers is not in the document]
xqft: Fixed!

Cargo.toml
xqft: The idea is to have multiple platforms in this crate for different gpus. Right now we only have Metal and in the coming days we'll probably have something with CUDA. We may need to reorg a bit like you suggested and have different modules for each platform. @mpaulucci @IAvecilla wdyt?

math/src/field/test_fields/u64_test_field.rs
mpaulucci: But this discussion is outside of the scope of this PR, so whatever, we can revert the changes here if it is easier.

Cargo.toml
IAvecilla: I like the idea of following the proving-system convention and creating a metal dir and declaring `gpu/metal` here. Right now CUDA is in early development so we can change the name to _metal_ if you think is a better approach

math/src/field/test_fields/u64_test_field.rs
xqft: Reverted the name change, after this is merged I'll open an issue with your suggestions as the `IsTwoAdicField` trait implements values which are valid only for a specific modulus.

jrchatruc (approved):

IAvecilla (approved):
LGTM!

gpu/src/fft/fft_metal.rs
```diff
@@ -0,0 +1,257 @@
...
+
+    pub fn execute_fft<F: IsTwoAdicField>(
+        &self,
+        input: &[FieldElement<F>],
+        (command_buffer, compute_encoder): (
+            &metal::CommandBufferRef,
+            &metal::ComputeCommandEncoderRef,
+        ),
+    ) -> Result<Vec<FieldElement<F>>, FFTError> {
+        let order = input.len().trailing_zeros() as u64;
```
ilitteri, line 77: Casts may panic, we could file an issue for replacing these `as` conversions with a new `FFTError`.

ilitteri: This way we ensure that if the code panics it'd be only because the `unsafe` block.

ilitteri (approved):

xqft:
Sorry to reviewers, I must mention here that this PR has a follow-up, #193, which pretends to clean up the messy bits of this POC before making it usable.

gpu/src/fft/fft_metal.rs
MauroToscano: This as won't panic, it's good, it's an u32 casted as an u64. Also, if it were the other way around, it wouldn't panic, it slices the value.

xqft: TIL, I agree, the main drive of these casts is because the `metal` crate seems to only handle `u64` integers, I assumed that a `usize` couldn't be bigger than 64 bits in most hardware but in any way we must change them.

MauroToscano (changes requested):
We need to remove metal dependency without features of Math, and add a default-workspace, else this won't compile in anything that's not a mac.

xqft:
Done, CI will now ignore the `gpu` crate, in our follow-up PR we'll introduce conditional compilation to make a better solution for this problem.

mpaulucci (approved):

PR #193, opened 2023-03-20
Implemented small Metal abstractions and make the FFT POC usable

xqft:
# Implemented small Metal abstractions and make the FFT POC usable

## Description

This is a follow-up for #190 aiming to structure the `gpu` crate and to implement a small set of Metal abstractions and a API that wraps around these for making the FFT POC usable. A second goal is to pave the way for future features, like parallel bit-reverse permutation, different FFT algorithms (orderings and radixes) and implementations in other platforms like CUDA.

## Type of change

- [x] New feature

## Checklist
- [ ] Unit tests added


Cargo.toml
```diff
@@ -4,4 +4,5 @@ members = [
     "math",
     "crypto",
     "proving-system/stark",
+    "gpu",
```
MauroToscano, line 7: Is it going to be shared between multiple GPUs ? Else I'd call this metal. Or metal GPU. Or it can be like proving-system, where you have a folder with multiple workspaces

gpu/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "lambdaworks-gpu"
```
MauroToscano, line 2: Is it going to be shared between multiple GPUs ? Else I'd call this lambdaworks-metal. Or lambdaworks-metal-gpu

math/src/field/test_fields/u64_test_field.rs
```diff
@@ -1,9 +1,9 @@
 use crate::field::traits::{IsField, IsPrimeField, IsTwoAdicField};
 
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct U64TestField<const MODULUS: u64>;
+pub struct U64Field<const MODULUS: u64>;
```
MauroToscano, line 4: Why this change ? There is an u64PrimeField for non testing cases. This part of the library should be used only for tests, not making that explicit may lead to confusion for users of the library. If you want to change the name wherever you use it, you can always use `as` wherever you import it

MauroToscano (changes requested):

MauroToscano:
I have added the same comments to the opened PR. These can be disregarded if fixed there PR #190

gpu/Cargo.toml
xqft: Discussed in #190

Cargo.toml
xqft: Discussed in #190

gpu/src/metal/shaders/fft.h.metal
```diff
@@ -0,0 +1,23 @@
+#pragma once
+#include "fp.h.metal"
+
+[[kernel]]
+void radix2_dit_butterfly(
+    device uint32_t* input [[ buffer(0) ]],
```
mpaulucci, line 6: you need to get latest `main` for the latest implementation

gpu/src/metal/shaders/twiddles.h.metal
```diff
@@ -0,0 +1,82 @@
...
+    uint
+);
+
+template [[ host_name("calc_twiddles_bitrev_inv") ]] 
+[[kernel]] void calc_twiddles_bitrev_inv<p256::Fp>(
+    device p256::Fp*,
+    constant p256::Fp&, 
+    uint,
+    uint
+);
```
IAvecilla, line 82: Add a new line here

IAvecilla (approved):

gpu/src/metal/fft/polynomial.rs
```diff
@@ -0,0 +1,145 @@
...
+    field::{
+        element::FieldElement,
+        traits::{IsTwoAdicField, RootsConfig},
+    },
+    polynomial::Polynomial,
+};
+
+use super::{errors::FFTMetalError, helpers::log2, ops::*};
+
+trait MetalFFTPoly<F: IsTwoAdicField> {
```
IAvecilla, line 12:
```suggestion
pub trait MetalFFTPoly<F: IsTwoAdicField> {
```

This trait should be public to expose these functions outside the `gpu` crate.

gpu/src/metal/fft/polynomial.rs
IAvecilla, line 1: We should match the interface of all the functions defined here with those declared on the `math` crate. I think it is better to keep the metal state as an internal structure and not ask for it in the API.

IAvecilla (changes requested):

mpaulucci (approved):

IAvecilla (approved):

MauroToscano (approved):

PR #198, opened 2023-03-21
Fully integrate with FFT API to interpolate/evaluate

### LINKED ISSUE #189: Fully integrate with FFT api to interpolate/evaluate
For now, we are only using it to generate the roots of the unit.

gabrielbosio:
## Description

Resolves #189.
Moves `air::constraints::helpers` module to `math`.

## Type of change

- [x] Optimization

## Checklist
- [x] Linked to Github Issue
- [x] Unit tests added


math/src/polynomial.rs
```diff
@@ -237,6 +237,32 @@ where
     Polynomial::interpolate(interpolation_points.as_slice(), values.as_slice())
 }
 
+pub fn compose_fft<F>(
```
mpaulucci, line 240: So, you have `poly_1` and `poly_2` and you want to calculate `poly_3 =  poly_1(poly_2)` right?

So the efficient way to do this with fft y to convert both `poly_1` and `poly_2` to value representation (by doing fft), then compose the values, and then interpolate (by doing inverse fft).

gabrielbosio: Makes sense, I only have one question: How do I compose value representations? Because the value representation of `poly_2` should be a vector of powers of a primitive root.

ilitteri (approved):

jrchatruc (approved):

mpaulucci (approved):

PR #209, opened 2023-03-23
Improve naming of CI jobs.

mpaulucci:
Changelog:
- Improved naming of CI jobs
- Merged fmt and clippy into lint.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,4 +1,4 @@
-name: Lambdaworks build checks
+name: CI
 on:
   push:
     branches: [ main ]
@@ -10,8 +10,8 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  check:
-    name: Check
+  build:
+    name: Build
     runs-on: ubuntu-latest
     steps:
       - name: Checkout sources
@@ -29,7 +29,8 @@ jobs:
         with:
           command: check
 
-  coverage:
+  test:
+    name: Test
     runs-on: macos-latest
     env:
       CARGO_TERM_COLOR: always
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -70,13 +48,27 @@ jobs:
...
         with:
-            toolchain: stable
-            components: clippy
-      - run: make clippy
+          toolchain: stable
+          components: clippy
+      - name: Install cargo-llvm-cov
+        uses: taiki-e/install-action@cargo-llvm-cov
+      - name: Generate code coverage
+        run: cargo llvm-cov --all-features --workspace --lcov --output-path lcov.info
```
IAvecilla, line 68: This step runs the tests and also generates the code coverage, maybe we can add that to the name, like `Run tests and generate code coverage`

IAvecilla (approved):
LGTM! some minor details

.github/workflows/ci.yaml
```diff
@@ -10,8 +10,8 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  check:
-    name: Check
+  build:
+    name: Build
```
ilitteri, line 14: This could be a little confusing since cargo check doesn't behave like a build ([it just compiles the packages without performing any code generation](https://doc.rust-lang.org/cargo/commands/cargo-check.html)).

ilitteri: Maybe we could use `compile` and `Compile`.

jrchatruc (approved):

.github/workflows/ci.yaml
mpaulucci: Sure, that works for me.

ilitteri (approved):

PR #210, opened 2023-03-23
Add support for `U256MontgomeryTwoAdicPrimeField` on Metal

mpaulucci:
## Description
Added support for a U256 Stark compatible field which will allow us to integrate the Metal FFT into the Stark prover.


math/src/field/element.rs
```diff
@@ -43,6 +43,18 @@ where
     }
 }
 
+impl<F> FieldElement<F>
```
mpaulucci, line 46: open to alternatives if this is not the right solution? The problem is we get the montgomery representation of the field back from metal and we need to store the `value` as-is when creating back the Field in Rust, instead of converting it to montgomery.

gpu/src/fft/fft_metal.rs
```diff
@@ -160,14 +160,10 @@ impl FFTMetalState {
         compute_encoder.set_buffer(0, Some(&omega_buffer), 0);
         compute_encoder.set_buffer(1, Some(&result_buffer), 0);
 
-        // SIMD group size:
```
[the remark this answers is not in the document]
xqft: yep I missed that

math/src/field/element.rs
MauroToscano: It looks good to me.

Maybe in the future we want to just make a big memory copy that sets up N elements, since we will have a huge amount of data and waiting for GPU is slow.

xqft (approved):
LGTM

IAvecilla (approved):

jrchatruc (approved):

Issue #34, opened 2023-02-09
Add `as_conversions` lint warning


ilitteri:
This issue involves:
- Adding `as_conversions` lint as a warning at a workspace level.
- Fixing the clippy warnings because of the above.

For more context see [as_conversions](https://rust-lang.github.io/rust-clippy/master/#as_conversions).

PR #212, opened 2023-03-25
As_conversions lint and corrections

tcoratger:
# Add `as_conversions` lint warning and corrections

## Description

As a response to #34, a `as_conversions` lint is added as a warning (`cargo clippy --all-targets --all-features -- -W clippy::as_conversions -D warnings`) and the proper corrections are implemented in the code to avoid all the `as` conversions and replace them by appropriate `try_from` methods.
- Changes implemented everywhere
- `#![allow(clippy::as_conversions)]` mention added for `pub const fn` where these kinds of fixs cannot be implemented (due to `cannot call non-const fn in constant functions` error).

## Type of change

- [x] Optimization

## Checklist
- [x] Linked to Github Issue


Makefile
```diff
@@ -16,4 +16,4 @@ benchmarks:
 
 # BENCHMARK should be one of the [[bench]] names in Cargo.toml
 benchmark:
-	cargo criterion --bench ${BENCH}
+	cargo criterion --bench ${BENCH}
```
jrchatruc, line 19: Add newline to the end of the file

jrchatruc:
Looks good other than the newline stuff, we can merge after that's fixed and merge conflicts are resolved

jrchatruc (approved):

tcoratger:
@jrchatruc Changes are done, can you merge as soon as you can? As this PR impacts lost of files, this will generate conflicts each time a new push is made on `main` branch :)

Makefile
```diff
@@ -4,7 +4,7 @@ test:
 	cargo test
 
 clippy:
-	cargo clippy --all-targets --all-features -- -D warnings
+	cargo clippy --all-targets --all-features -- -W clippy::as_conversions -D warnings
```
ilitteri, line 7: We should better add the lint in the `lib.rs` files (for each lib in the workspace). The idea is to add more lints in the future; if we add them to this command, it will become too large.

ilitteri (changes requested):

ilitteri:
> @jrchatruc @ilitteri I'm not sure that this error `Error: Codecov: Failed to properly upload: The process '/Users/runner/work/_actions/codecov/codecov-action/v3/dist/codecov' failed with exit code 255` is related to the PR?

I've ran the job again and it seems to be ok, we could merge after solving the conflict.

ilitteri (approved):

tcoratger:
@ilitteri No problem, you can just let me know the day you want to merge, so that I resolve the conflicts quickly and you merge in stride (to avoid a new push on main which would lead to new conflicts to be resolved)

ilitteri:
Hello again, sorry for the delay! First of all thanks for the contribution. We’ve been talking with the team and we agreed that this PR is still to be polished. The main idea of removing the as conversions was because they could make the code panic, and simply replacing them with try_into/try_from and unwrap doesn’t solve this. It is ok to use try_into/try_from though, but to wrap up this PR we’d need to handle the errors instead of panicking explicitly.

ilitteri (changes requested):
Hello again, sorry for the delay! First of all thanks for the contribution. We’ve been talking with the team and we agreed that this PR is still to be polished. The main idea of removing the as conversions was because they could make the code panic, and simply replacing them with try_into/try_from and unwrap doesn’t solve this. It is ok to use try_into/try_from though, but to wrap up this PR we’d need to handle the errors instead of panicking explicitly.

tcoratger:
Hello @ilitteri no problem, as a crypto researcher, such a good repo to learn ZKP and crypto stuffs :) 
Sorry, I was thinking that it was just a `lint` stuff to upgrade the repo and then being able to play case by case. So, if I understand well, you are looking for something like that for each case right?

```rust
let x = match u64::try_from(2_u128.pow(100)) {
    Ok(val) => val,
    Err(err) => {
        eprintln!("Error: {}", err);
        return;
    }
};
```

crypto/src/commitments/kzg.rs
```diff
@@ -135,7 +135,11 @@ mod tests {
         let g2 = BLS12381TwistCurve::generator();
         let powers_main_group: Vec<G1> = (0..100)
```
Oppen, line 136:
```suggestion
        let powers_main_group: Vec<G1> = (0..100u128)
```

crypto/src/commitments/kzg.rs
```diff
@@ -135,7 +135,11 @@ mod tests {
...
             .map(|exponent| {
-                g1.operate_with_self(toxic_waste.pow(exponent as u128).representative())
+                g1.operate_with_self(
+                    toxic_waste
+                        .pow(u128::try_from(exponent).unwrap())
+                        .representative(),
+                )
```
Oppen, line 142:
```suggestion
                g1.operate_with_self(toxic_waste.pow(exponent).representative())
```

math/benches/benchmarks/fft.rs
```diff
@@ -32,7 +32,9 @@ pub fn fft_benchmarks(c: &mut Criterion) {
             |bench, coeffs| {
                 bench.iter(|| {
                     let mut coeffs = coeffs.clone();
-                    let twiddles = F::get_twiddles(pow as u64, RootsConfig::BitReverse).unwrap();
```
Oppen, line 35: Similarly to the other comments, just create the range from `u64`.

math/benches/benchmarks/fft.rs
```diff
@@ -43,7 +45,8 @@ pub fn fft_benchmarks(c: &mut Criterion) {
             |bench, coeffs| {
                 bench.iter(|| {
                     let mut coeffs = coeffs.clone();
-                    let twiddles = F::get_twiddles(pow as u64, RootsConfig::Natural).unwrap();
+                    let twiddles =
+                        F::get_twiddles(u64::try_from(pow).unwrap(), RootsConfig::Natural).unwrap();
```
Oppen, line 49: And here as well.

math/src/elliptic_curve/point.rs
```diff
@@ -108,7 +108,8 @@ mod tests {
     #[test]
     fn operate_with_self_works_2() {
         let mut point_1 = TestCurve1::generator();
-        point_1 = point_1.operate_with_self(TEST_CURVE_1_MAIN_SUBGROUP_ORDER as u128);
+        point_1 =
+            point_1.operate_with_self(u128::try_from(TEST_CURVE_1_MAIN_SUBGROUP_ORDER).unwrap());
```
Oppen, line 112: Consider making the constant `u128` instead.

math/src/fft/abstractions.rs
```diff
@@ -55,7 +55,8 @@ pub fn inverse_fft<F: IsTwoAdicField>(
     in_place_bit_reverse_permute(&mut results);
 
     for elem in &mut results {
-        *elem = elem.clone() / FieldElement::from(coeffs.len() as u64); // required for inverting the DFT matrix.
+        *elem = elem.clone() / FieldElement::from(u64::try_from(coeffs.len()).unwrap());
```
Oppen, line 58: I think you'll find `usize` often enough to merit a `From` implementation instead.

Oppen:
I'm not too sure about these changes. Most uses here are for promoting unsigned types to wider variants, which is entirely safe, and the counterpart is making the code extremely verbose. I would also worry about adding too much branching in elementary operations, that may harm performance.
Is there any other alternative? Can the lint be made a little more lax so it only complains about conversion between signed and unsigned and from a wider to a narrower type? Those are the actually risky ones.

Oppen:
> Hello @ilitteri no problem, as a crypto researcher, such a good repo to learn ZKP and crypto stuffs :) Sorry, I was thinking that it was just a `lint` stuff to upgrade the repo and then being able to play case by case. So, if I understand well, you are looking for something like that for each case right?
> 
> ```rust
> let x = match u64::try_from(2_u128.pow(100)) {
>     Ok(val) => val,
>     Err(err) => {
>         eprintln!("Error: {}", err);
>         return;
>     }
> };
> ```

Sort of. You would need to alter signatures to bubble errors up. But that again would add overhead to the whole stack. Besides, the branching would probably make non-constant-time all the code, which means it would also make it so for places susceptible to timing attacks, right?

MauroToscano:
This hasn't been active for a while, and it has many conflicts. I'm closing it. We can re open it if needed.

PR #215, opened 2023-03-28
Stark implementation docs

jrchatruc:
## Type of change

Please delete options that are not relevan

- [x] Documentation


proving_system/stark/stark_docs/src/recap.md
diegokingston, line 1: Line 181 lacks the random coefficients for the linear combination. 
Perhaps in line 195 we should say that we construct a Merkle tree from the evaluations of the LDE

jrchatruc: The first point was addressed in [86f8b50](https://github.com/lambdaclass/lambdaworks/pull/215/commits/86f8b507094bb895fdd5083d30c5ebe9f86897c2).

For the second one we more generally need to add the consistency check for the DEEP composition polynomial. I created an issue for it [here](https://github.com/lambdaclass/lambdaworks/issues/224).

PR #218, opened 2023-03-28
Fix degree calculations for composition polynomial

gabrielbosio:
## Description

- The degree of `roots_of_unity_vanishing_polynomial` used to calculate AIR transition divisors was calculated using log2 of the trace length instead of using simply the trace length.
- Substracted max degree used for the composition polynomial with zerofier degree.
- Calculated transition constraint zerofier degree using trace length.

## Type of change

- [x] Bug fix


entropidelic (approved):

pablodeymo (approved):
LGTM

MauroToscano (approved):

PR #225, opened 2023-03-30
Cairo VM instructions decoding

MauroToscano:
# Decode instructions from the memory

## Description

Instructions needs to be decoded and joined with the Cairo trace to make a "Pre Trace" for the cairo Prover


proving_system/stark/src/cairo_vm/cairo_instruction.rs
```diff
@@ -0,0 +1,545 @@
...
+#[derive(Clone, Debug, PartialEq)]
+pub struct CairoInstructionFlags {
+    pub opcode: CairoOpcode,
+    pub ap_update: ApUpdate,
+    pub pc_update: PcUpdate,
+    pub res_logic: ResLogic,
+    pub op1_src: Op1Src,
+    pub op0_reg: Op0Reg,
+    pub dst_reg: DstReg,
+}
```
MauroToscano, line 185: I'd put this at the beggining since it's the important thing, along with the rest of the instruction

MauroToscano (commented):
This is just the flags, we need more than this, but it's good!

We could add a commentary on the tests telling what the hexstrings mean, or from where did we got them, so others can generate more in the feature or validate the test cases better

proving_system/stark/src/cairo_vm/cairo_instruction.rs
```diff
@@ -0,0 +1,545 @@
+/*
+   TO DO: Find a better name
```
MauroToscano, line 2: Oh, we should remove this, or change the name. I think we can add the rest of the decode and see how we want to organize this better. I'm thinking having two files may be useful, since the code will become longer

entropidelic:
I have added some reference in the tests about the instruction decoding. The instructions were obtained with cairo-rs

pablodeymo (approved):
LGTM

entropidelic (approved):

PR #227, opened 2023-03-30
PyO3 field element

mfachal:
# Field Element implementation

## Description

Add FieldElement (u256) wrapper

## Type of change

- [ ] New feature
- [ ] Bug fix
- [ ] Optimization

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run


py/.github/workflows/CI.yml
```diff
@@ -0,0 +1,118 @@
+# This file is autogenerated by maturin v0.14.15
```
gabrielbosio, line 1: I don't think this workflow is considered in Github Actions, we can remove/ignore it.

py/src/merkle_tree/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod proof;
```
gabrielbosio, line 1: Why `merkle_tree` module is needed to wrap `FieldElement`?

py/src/merkle_tree/proof.rs
```diff
@@ -0,0 +1,7 @@
+use pyo3::prelude::*;
+
+#[pyclass(name = "Proof")]
+pub struct PyProof {
```
gabrielbosio, line 4: Why `PyProof` is needed to wrap `FieldElement`?

py/src/proving_systems/stark/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+    pub fn inv(&self) -> Self {
+        Self(self.0.inv())
+    }
+
+    pub fn pow(&self, pyexp: &PyInt) -> PyResult<Self> {
+        let exp: u64 = pyexp.extract()?;
+        Ok(Self(self.0.pow(exp)))
+    }
+
+    pub fn __pow__(&self, pyexp: &PyInt, modulo: Option<&PyInt>) -> PyResult<Py<PyAny>> {
```
gabrielbosio, line 66: Why this function is public when `__add__`, `__sub__`, etc. are private?

py/src/proving_systems/stark/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+    pub fn zero() -> Self {
+        Self(FE::zero())
+    }
+}
+
+#[pyclass(name = "StarkProofConfig")]
+pub struct PyProofConfig(pub ProofConfig);
+
+#[pymethods]
+impl PyProofConfig {
```
gabrielbosio, line 90: I get why `PyProof` would be necessary. We can remove `prove` and `verify` because this PR is not adding wrappers for proving and verifying.

py/tests/lamdaworks_py_test.py
```diff
@@ -0,0 +1,83 @@
...
+class TestFibonacci(unittest.TestCase):
+    @classmethod
+    def _fibonacci_trace(cls, v0, v1):
+        N = 32
+        l = [v0, v1]
+        for i in range(2, N):
+            l.append(l[i - 1] + l[i - 2])
+        return l
+
+    def test_prove(self):
```
gabrielbosio, line 73: We can remove this test if we remove the `prove` and `verify` methods.

py/src/proving_systems/stark/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+            count_queries,
+            blowup_factor,
+        })
+    }
+}
+
+#[pyclass]
+pub struct PyStarkProof(StarkProof<PrimeField>);
+
+// #[pyfunction]
```
gabrielbosio, line 103: We can remove this commented method.

py/src/proving_systems/stark/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+//     };
+
+//     Ok(PyStarkProof(lambdaworks_stark::prover::prove(
+//         &trace,
+//         &proof_config.0,
+//     )))
+// }
+
+#[pyfunction]
+pub fn verify(_stark_proof: &PyStarkProof) -> bool {
```
gabrielbosio, line 122: We can remove this function

gabrielbosio (changes requested):
Looks good, I think we can change some stuff

gabrielbosio (approved):

MauroToscano (commented):
I'm not seeing on tests what happens in the case of errors, nor documentation. So i'd add that. Ex: Division by zero

It would be nice to use only one Makefile in the project, since that's what all the modules are doing.

It would be nice to have documentation on how to run, which python version do you have to use, and that kind of stuff. It can be on the readme of the library so it's easy to find.

jrchatruc (approved):

gabrielbosio (dismissed):
Before merging this, we should add the corresponding test step to the CI workflow. Related issue: #250

Makefile
```diff
@@ -1,7 +1,13 @@
...
 
-test:
+py.develop:
+	python3 -m venv .venv
+	. .venv/bin/activate
+	maturin develop -m py/Cargo.toml
+
+test: py.develop
 	cargo test
+	python3 -m unittest py/tests/lamdaworks_py_test.py
```
Oppen, line 10: Invert the order of these. The first target is the default target in `make`, which would be expected to be either `test` or `build`.

py/src/proving_systems/stark/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+    fn __sub__(&self, other: &Self) -> Self {
+        Self(&self.0 - &other.0)
+    }
+
+    fn __mul__(&self, other: &Self) -> Self {
+        Self(&self.0 * &other.0)
+    }
+
+    fn __truediv__(&self, other: &Self) -> PyResult<Self> {
+        match other.0 == FE::zero() {
```
Oppen, line 43: I'm assuming the following method exists:
```suggestion
        match other.0.is_zero() {
```

py/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "lambdaworks_py"
+version = "0.1.0"
+edition = "2021"
```
Oppen, line 4: It may be a good idea to define these at the package level and inherit it in the crates.

py/src/proving_systems/stark/mod.rs
gabrielbosio: I didn't found that method. Maybe we can implement it in the future.

py/Cargo.toml
gabrielbosio: We added those lines to keep consistency with the other crates. However, we can change this in another PR.

MauroToscano:
This wrapper is no longer maintained. We can re open it if we take this feature again and see how it goes.

PR #230, opened 2023-03-31
add diferential fuzzer plonk vs gnark

daphneherlambda:
# differential fuzzer for plonk vs gnark

## Description

this PR adds a fuzzer that compares the plonk implementation vs the gnark plonk implementation

## the structure

* in gnark_backend there is a main.go file where gnark code is imported and the functions we are going to use are defined
* there´s a build.rs file used to compile go code when running the rust program in order to use it
* there´s a Makefile that does the building
* in src folder we have the folder implementation, inside we use the go code exported as C also we have the backend helper modules

as is currently a work in progress theres still missing the logic in main.go that gets a circuit and a witness and returns the verified result

### DIFF
--- a/fuzzer/Cargo.toml
+++ b/fuzzer/Cargo.toml
@@ -1,0 +1,9 @@
+[package]
+name = "fuzzer"
+version = "0.1.0"
+edition = "2021"
+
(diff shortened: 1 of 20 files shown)


MauroToscano:
This has been inactive for really long

PR #233, opened 2023-04-03
[STARKs] Add consistency check for the DEEP composition polynomial

### LINKED ISSUE #224: [STARKs] Add consistency check for the DEEP composition polynomial
Right now all the prover does with the DEEP composition polynomial is commit to it using `FRI`, but there is no check of the relationship of this polynomial with the trace or the composition polynomial `H` used before. 

For this, a consistency check needs to happen where the prover first commits to the trace `LDE` using a merkle tree. Then, before doing `FRI` on `DEEP(x)`, a random value in the `LDE` coset domain is sampled and the prover provides both `DEEP(x_0)` and the trace evaluations `t(x_0)`, along with all the others needed to reconstruct `DEEP(x_0)` on the verifier side.

The verifier then does the usual thing of reconstructing `DEEP(x_0)` and checking that it matches the claimed evaluation from the prover.

Once this is added, update the prover documentation accordingly to include this step.

gabrielbosio:
## Description

Resolves #224

## Type of change

- [x] New feature

## Checklist
- [x] Linked to Github Issue
- [x] Verification of elements of DEEP composition polynomial as trace evaluations
- [x] Verification of DEEP composition polynomial correctness


proving_system/stark/src/prover.rs
```diff
@@ -4,231 +4,331 @@ use super::{
...
-            poly.evaluate_offset_fft(
-                &FieldElement::<F>::from(air.options().coset_offset),
-                air.options().blowup_factor as usize,
-            )
-        })
-        .collect::<Result<Vec<Vec<FieldElement<F>>>, FFTError>>()
+pub struct Prover<'a, F: IsTwoAdicField, A: AIR<Field = F>> {
+    air: &'a A,
+    transcript: Transcript,
+}
```
jrchatruc, line 25: Let's revert this `Prover` struct since having the transcript as a field can get very error prone, as we would need to make sure to reset the transcript's state if new calls to prove are made, etc.

gabrielbosio:
Instead of adding `Prover` and `Verifier` structs, I will add a struct for each function that has more than 7 arguments.

proving_system/stark/src/prover.rs
gabrielbosio: Done

docs/src/proving_systems/starks/api.md
```diff
@@ -72,14 +68,14 @@ The way we specify our fibonacci transition constraint looks like this:
...
-    ) -> Vec<FieldElement<Self::Field>> {
+    &self,
+    frame: &air::frame::Frame<Self::Field>,
+) -> Vec<FieldElement<Self::Field>> {
     let first_row = frame.get_row(0);
     let second_row = frame.get_row(1);
     let third_row = frame.get_row(2);
 
-    vec![third_row[0] - second_row[0] - first_row[0]]
+    vec![third_row[0].clone() - second_row[0].clone() - first_row[0].clone()]
```
MauroToscano, line 78: This is the compute of quadratic constraints, right ? But the documentation. It can be changed for 

`vec![&second_row[0] - &first_row[0] * &first_row[0]]`

No need for clones

gabrielbosio: I removed the `clone`s, Thanks!

MauroToscano (approved):

proving_system/stark/src/prover.rs
```diff
@@ -158,11 +157,28 @@ where
...
+        &composition_poly_odd,
+    );
+
     let query_list = (0..air.context().options.fri_number_of_queries)
-        .map(|_| {
-            // * Sample q_1, ..., q_m using Fiat-Shamir
-            let q_i = transcript_to_usize(transcript) % 2_usize.pow(lde_root_order);
-            transcript.append(&q_i.to_be_bytes());
+        .enumerate()
+        .map(|(i, _)| {
```
entropidelic, line 173: I think te `enumerate()` can be removed and use the index directly from the `(0..air.context().options.fri_number_of_queries)` range

schouhy (approved):

entropidelic (approved):

PR #239, opened 2023-04-10
Benchmarks for stark prover

jrchatruc:
## Description

This PR adds benchmarks for the STARK prover. Because we don't yet have a way to specify `AIR`s through files, I moved our test `AIR` to a separate shared module so both tests and benchmarks can use them. Eventually they should all live as files in something like an `examples` directory.

Benchmarks are also deployed to gh-pages under the `/bench` path, so once this is merged they'll be on https://lambdaclass.github.io/lambdaworks/bench

## Type of change

- [x] New feature


math/benches/benchmarks/fft.rs
```diff
@@ -21,7 +21,7 @@ pub fn fft_benchmarks(c: &mut Criterion) {
     let mut group = c.benchmark_group("fft");
     group.sample_size(10); // it becomes too slow with the default of 100
 
-    for pow in 20..=24 {
+    for pow in 20..22 {
```
jrchatruc, line 24: This was lowered because the benchmarks take way too long

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,36 @@
+name: benchmarks 
+
+on:
+  push:
+    branches: [ main ]
+
+permissions:
+  # deployments permission to deploy GitHub pages website
```
mpaulucci, line 8: comment seems unrelated?

mpaulucci (approved):

PR #240, opened 2023-04-10
Remove `new` constructor from AIR trait

jpcenteno:
# Remove the `new` constructor method from the `AIR` trait.

## Description

We need to add more parameters to our `AIR` implementation, but the trait only allows for a constructor with a `AirContext` parameter. I removed the constructor from the trait since it's not used by other functions, and updated the tests that used it by replacing the `new` method with a `From<AirContext>` implementation.

## Type of change

- Refactor

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run

### DIFF
--- a/proving_system/stark/benches/benchmarks/stark.rs
+++ b/proving_system/stark/benches/benchmarks/stark.rs
@@ -6,7 +6,6 @@ use lambdaworks_stark::{
     air::{
         context::{AirContext, ProofOptions},
         trace::TraceTable,
-        AIR,
     },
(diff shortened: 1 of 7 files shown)


jrchatruc (approved):

Oppen (approved):

MauroToscano (approved):

MauroToscano:
It's good, traits in general do not need a new, I did the same for hashers. But we have conflicts on it now

PR #242, opened 2023-04-10
Add feature flag to disable Fiat-Shamir

jrchatruc:
## Description

This PR adds a feature flag `test_fiat_shamir` that effectively disables Fiat-Shamir when running the stark prover and verifier. This means that if you run something like

```
cargo test test_prove_fib --features test_fiat_shamir
```

the fibonacci tests will run with a transcript that always returns the same value. This is useful for debugging.

To do this, `Transcript` was turned into a `Trait` and a `DefaultTranscript` and `TestTranscript` structs were introduced.

## Type of change

- [x] New feature


MauroToscano:
In the future if we add more debug tools like this we could add one feature to enable them all, for simplicity

MauroToscano (changes requested):

MauroToscano (approved):

entropidelic (approved):

PR #244, opened 2023-04-11
Update hash in merkle tree to SHA3 256

MauroToscano:
# Update hash function in Merkle Tree

## Description

Hash functions needs to be updated for a 
This PR includes:
- Change of the hash function used in Merkle Tree from a Test One to SHA 3 256
- Fix a bug where  Merkle Proof verification wasn't working  properly 
- Fix a bug where if the same element was duplicated in the Merkle Tree, and by using Fiat Shamir one that's not the first was chosen, Stark Prover didn't work
- Fix a bug where stark verifier was calculating the symmetric index wrongly as ` let layer_evaluation_index_symmetric =
            (q_i + current_layer_domain_length) % current_layer_domain_length;` instead of `let layer_evaluation_index_symmetric =
            (q_i + current_layer_domain_length / 2) % current_layer_domain_length;`, so the proof wouldn't verify
- Remove `get_proof` function from Merkle, since it's behaviour was strange when duplicated where included, and it wasn't needed. 
- Refactor the Merkle Tree it's no longer generalized over the hasher, and instead one is provided when it's created and to do the verification, to simplify the code
- Remove `new()` from `CryptoHash` trait, since new is always called for specific instances. This allows the specific implementation to have a constant `new()` which is useful for initializing the hasher, since the hasher can then be a constant for the program

After merging this we need to either implement a better hash to curve algorithm, see https://www.ietf.org/id/draft-irtf-cfrg-hash-to-curve-16.html, or change the MerkleTree to work with binary data, to improve security.

## Type of change

- [ x ] New feature
- [ x ] Bug fix


crypto/src/hash/poseidon/mod.rs
```diff
@@ -15,6 +15,21 @@ pub struct Poseidon<F: IsField> {
     params: Parameters<F>,
 }
 
+impl Poseidon<BLS12381PrimeField> {
+    pub fn new() -> Self {
+        Self {
+            params: Parameters::with_t2()
+                .expect("Error loading parameters for Posedon BLS12381 hasher"),
```
Oppen, line 22: Is this a "can't logically happen" situation?

Oppen: And if so, is it an option to make it infallible instead?

crypto/src/merkle_tree/proof.rs
```diff
@@ -11,25 +11,28 @@ use lambdaworks_math::{
...
         mut index: usize,
         value: &FieldElement<F>,
-    ) -> bool {
-        let mut hashed_value = self.hasher.hash_one(value.clone());
+        hasher: &dyn IsCryptoHash<F>,
+    ) -> bool
+    where
+        FieldElement<F>: ByteConversion,
+    {
+        let mut hashed_value = hasher.hash_one(value.clone());
```
Oppen, line 29: I find it odd that the hashers need to take ownership of their arguments. Why is that?

math/src/unsigned_integer/element.rs
```diff
@@ -477,15 +477,18 @@ impl<const NUM_LIMBS: usize> ByteConversion for UnsignedInteger<NUM_LIMBS> {
...
-            let limb = u64::from_be_bytes(
-                chunk
-                    .try_into()
-                    .map_err(|_| ByteConversionError::FromBEBytesError)?,
-            );
-            limbs.push(limb);
-            Ok::<_, ByteConversionError>(())
-        })?;
+
+        bytes[0..NUM_LIMBS * 8]
```
Oppen, line 481: This could panic if the argument doesn't have enough whole limbs. Please consider using `bytes.get(0..NUM_LIMBS * 8)?` instead, as this allows catching the failure gracefully.

math/src/unsigned_integer/element.rs
```diff
@@ -496,15 +499,18 @@ impl<const NUM_LIMBS: usize> ByteConversion for UnsignedInteger<NUM_LIMBS> {
...
-        bytes.chunks_exact(8).rev().try_for_each(|chunk| {
-            let limb = u64::from_le_bytes(
-                chunk
-                    .try_into()
-                    .map_err(|_| ByteConversionError::FromLEBytesError)?,
-            );
-            limbs.push(limb);
-            Ok::<_, ByteConversionError>(())
-        })?;
+        bytes[0..NUM_LIMBS * 8]
```
Oppen, line 502: Ditto.

crypto/src/merkle_tree/proof.rs
MauroToscano: You're right. Taking ownership should be pointless. I'll make the change and see how it goes

math/src/unsigned_integer/element.rs
MauroToscano: Makes sense

mpaulucci (approved):
lgtm

crypto/src/hash/poseidon/mod.rs
Oppen: Cool then.

math/src/unsigned_integer/element.rs
```diff
@@ -478,17 +478,26 @@ impl<const NUM_LIMBS: usize> ByteConversion for UnsignedInteger<NUM_LIMBS> {
...
+        // We cut off extra bytes, this is useful when you use this function to generate the element from randomness
+        // In the future with the right algorithm this shouldn't be needed
+
+        let needed_bytes_res = bytes.get(0..NUM_LIMBS * 8);
+
+        if needed_bytes_res.is_none() {
+            return Err(ByteConversionError::FromBEBytesError);
+        }
+
+        let needed_bytes = needed_bytes_res.unwrap();
```
Oppen, line 490: nit(non-blocking): you can make this more brief as follows:
```suggestion
        let needed_bytes = bytes.get(0..NUM_LIMBS * 8)
            .ok_or(ByteConversionError::FromBEBytesError)?;
```

Oppen: This way you could also keep the chain of composition if you want.

Oppen (approved):

PR #245, opened 2023-04-11
Added polynomial interpolation benchs and markdown table comparison, and other improvements

xqft:
# Added polynomial interpolation benchs and markdown table comparison, and other improvements

## Description

This adds benchmarks for comparing different methods of polynomial interpolation and uploaded local results to a markdown table. Sample size has been reduced because of local performance issues.

Also made different improvements on existing benchmarks, like better identifiers, so criterion can make better reports about them.

## Type of change

- [ ] New feature
- [ ] Bug fix


gpu/benches/benchmarks/metal.rs
```diff
@@ -1,41 +1,49 @@
...
                     // TODO: autoreleaspool hurts perf. by 2-3%. Search for an alternative
                     objc::rc::autoreleasepool(|| {
                         let coeffs = coeffs.clone();
                         let metal_state = MetalState::new(None).unwrap();
-                        let twiddles =
-                            F::get_twiddles(order as u64, RootsConfig::BitReverse).unwrap();
+                        let twiddles = F::get_twiddles(order, RootsConfig::BitReverse).unwrap();
 
                         fft(&coeffs, &twiddles, &metal_state).unwrap();
                     });
```
IAvecilla, line 49: We can use the function to evaluate directly from the polynomial to be the same as the benchmark for interpolation. Create a polynomial from the coeffs before the benchmark starts and measure `poly.evaluate_fft_metal(&metal_state)` that already has the twiddle generation.

IAvecilla (approved):

gpu/README.md
```diff
@@ -0,0 +1,25 @@
+## Polynomial interpolation methods comparison
```
mpaulucci, line 1: shouldn't we use the /docs folder?

gpu/README.md
```diff
@@ -0,0 +1,25 @@
...
+- **CPU FFT:** Finding the lowest degree polynomial that interpolates pairs of twiddle factors and Fourier coefficients (the results of applying the Fourier transform to the coefficients of a polynomial) (see `math/src/polynomial.rs:interpolate_fft()`).
+- **GPU FFT (Metal):** Same as CPU FFT but the FFT algorithm is run on the GPU via the Metal API (Apple).
+
+these were run with criterion-rs in a MacBook Pro M1 (18.3), statistically measuring the total run time of one iteration. The field used was a 256 bit STARK-friendly prime field.
+
+All values of time are in milliseconds. Those cases which were greater than 30 seconds were marked respectively as they're too slow and weren't worth to be benchmarked. The input size refers to *d + 1* where *d* is the polynomial's degree (so size is amount of coefficients).
+
+| Input size | CPU Lagrange | CPU FFT   | GPU FFT (Metal) |
+|------------|--------------|-----------|-----------------|
+| 2^4        | 2.1973 ms    | 0.2104 ms | 2.5050 ms       |
```
mpaulucci, line 15: nit, I would use 1 precision number, i.e 2.5 ms instead of 2.5050 ms

gpu/benches/benchmarks/metal.rs
```diff
@@ -1,41 +1,49 @@
...
     field::{element::FieldElement, traits::IsTwoAdicField},
-    field::{test_fields::u32_test_field::U32TestField, traits::RootsConfig},
+    field::{fields::fft_friendly::stark_252_prime_field::Stark252PrimeField, traits::RootsConfig},
+    polynomial::Polynomial,
+    unsigned_integer::element::UnsignedInteger,
 };
 use rand::random;
 
-type F = U32TestField;
+type F = Stark252PrimeField;
```
mpaulucci, line 14: nice!

mpaulucci (approved):
+1 after comments are addressed

gpu/benches/benchmarks/metal.rs
xqft: I would propose to add a different benchmark for polynomial evaluation, and strip this one from twiddle gen and leave only ordered FFT, so each implementation (Metal, sequential and in the near future CUDA and WebGPU) can be compared easily. wdyt?

gpu/README.md
xqft: Done!

gpu/benches/benchmarks/metal.rs
IAvecilla: Yeah, seems like a good idea

jrchatruc (approved):

gpu/benches/benchmarks/metal.rs
xqft: Done!

PR #252, opened 2023-04-13
Extract FFT code into its own crate

mpaulucci:
# Extract FFT into its own crate

## Description
We want to avoid circular dependencies between `math` and `gpu` crates, so we decided to extract all the fft functionality into it's own module.

This way:
- `gpu` depends on `math`
- `fft` depends on `math` and `gpu` (if metal or cuda feature flag are on)

## Type of change
- [x] Refactor


gpu/benches/benchmarks/metal.rs
```diff
@@ -35,7 +35,8 @@ pub fn metal_fft_benchmarks(c: &mut Criterion) {
                         let coeffs = coeffs.clone();
                         let metal_state = MetalState::new(None).unwrap();
                         let twiddles =
-                            F::get_twiddles(order as u64, RootsConfig::BitReverse).unwrap();
+                            gen_twiddles(order as u64, RootsConfig::BitReverse, &metal_state)
```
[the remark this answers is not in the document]
xqft: That was a bug actually so it's good, though I already fixed it [here](https://github.com/lambdaclass/lambdaworks/pull/245/files#diff-78c6dc3bedd9137ff937c0cb21bb2a801f826b0119dc08f2504159e2695cd27eR36)

fft/src/roots_of_unity.rs
```diff
@@ -0,0 +1,69 @@
+use lambdaworks_math::field::{
```
mpaulucci, line 1: this was moved out of the TwoAdicity trait and into it's own module inside `fft`.

gpu/src/metal/fft/polynomial.rs
```diff
@@ -43,103 +38,23 @@ impl<F: IsTwoAdicField> MetalFFTPoly<F> for Polynomial<FieldElement<F>> {
...
+        let twiddles = gen_twiddles(order, RootsConfig::BitReverseInversed, &metal_state)?;
 
-        let coeffs = fft(fft_evals, &twiddles, state)?;
+        let coeffs = fft(fft_evals, &twiddles, &metal_state)?;
 
         let scale_factor = FieldElement::from(fft_evals.len() as u64).inv();
         Ok(Polynomial::new(&coeffs).scale_coeffs(&scale_factor))
     }
 }
-
```
mpaulucci, line 67: moved to `fft/src/polynomial`

math/src/field/traits.rs
```diff
@@ -29,64 +26,26 @@ pub trait IsTwoAdicField: IsField {
...
+            return Err(FieldError::RootOfUnityError(
+                "Order cannot exceed 2^{F::TWO_ADICITY}".to_string(),
                 order,
             ));
         }
-        let power = 1u64 << (Self::TWO_ADICITY - order);
+        let power = 1u64 << (F::TWO_ADICITY - order);
         Ok(two_adic_primitive_root_of_unity.pow(power))
     }
-
```
mpaulucci, line 50: moved to `roots_of_unity`

xqft (approved):

fft/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "lambdaworks-fft"
+version = "0.1.0"
+edition = "2021"
```
Oppen, line 4: From previous bad experiences, I strongly recommend you set a workspace level version and make these use the workspace one.
See https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table

fft/src/errors.rs
```diff
@@ -16,3 +16,14 @@ pub enum FFTError {
     #[error("Couldn't create a new Metal compute pipeline: {0}")]
     MetalPipelineError(String),
 }
+
+impl From<FieldError> for FFTError {
+    fn from(error: FieldError) -> Self {
+        match error {
+            FieldError::DivisionByZero => {
+                panic!("Can't divide by zero during FFT");
```
Oppen, line 24: Please try not to panic in library crates. You should always give our downstream users the opportunity to handle failure.

Oppen (approved):
LGTM, but check the **non-blocking** comments if you have time.

.rusty-hook.toml
```diff
@@ -1,5 +1,5 @@
 [hooks]
-pre-commit = "cargo test && cargo clippy --all-targets --all-features -- -D warnings && cargo fmt --all -- --check"
+pre-commit = "cargo test && cargo clippy --all-targets -- -D warnings && cargo fmt --all -- --check"
```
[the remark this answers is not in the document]
mpaulucci: Because of this: https://github.com/lambdaclass/lambdaworks/blob/main/gpu/src/lib.rs#L1-L5

PR #254, opened 2023-04-13
Use FFT interpolation in Plonk proving system

IAvecilla:
# Use FFT interpolation in Plonk proving system

## Description

To speed up all the rounds in the Plonk proving system, it's better to use all the operations implemented in FFT instead of Lagrange interpolation. I implemented the `IsTwoAdicField` trait to the `ScalarField` of the BLS12-381 curve to make it suitable to use with FFT. These changes affect the proving and the verifying function.

## Type of change
- [x] Optimization

## Checklist
- [ ] Linked to Github Issue
  - [ ] Benchmarks added/run


MauroToscano:
No longer valid

Issue #249, opened 2023-04-12
Port benchmarks to Iai


xqft:
The [criterion FAQ](https://github.com/bheisler/criterion.rs/blob/master/book/src/faq.md) discourages its use on CI workflows because of unreliable results (I got [notified](https://github.com/lambdaclass/lambdaworks/commit/bf0158c34593ccfc22ac7ef2ed236a950fbcfc2d#commitcomment-108518426) of a perf. regression with ratio up to 3.25 of code I didn't touch), so porting all benchmarks to [iai](https://bheisler.github.io/criterion.rs/book/iai/getting_started.html) seems necessary.

PR #255, opened 2023-04-14
Add iai support and benchmarks for performance comparison between new PRs and main

xqft:
# Add iai support and benchmarks for performance comparison between new PRs and main

## Description

This PR refactors all benchmarks for adding support for iai so it can be used to detect regressions in the CI. Two new CI jobs were added for this purpose. Now on every PR iai will execute some benchmarks and compare them to the main branch. This PR was made taking [cairo-rs](https://github.com/lambdaclass/cairo-rs) as reference.

This also includes a small refactor of jobs, and including Apple sillicon benches given the recent merge of #266.

This will solve #249 .

**Note for reviewers**: the iai jobs will fail because the main branch tests are not named the same as in this PR. After merging this should solve itself.

## Type of change

- [x] Benchmarks


xqft:
Note for reviewers: the iai jobs will fail because the main branch tests are not named the same as in this PR. After merging this should solve itself.

.github/workflows/criterion_benchs.yml
```diff
@@ -1,10 +1,8 @@
-name: Benchmarks 
+name: CI
```
mpaulucci, line 1: why this change?

.github/workflows/iai_benchs_main.yml
```diff
@@ -0,0 +1,34 @@
+name: CI
```
mpaulucci, line 1: why two separate files for benchmark jobs?

.github/workflows/criterion_benchs.yml
xqft: To be consistent with the rest of the CI checks

.github/workflows/criterion_benchs.yml
```diff
@@ -23,7 +22,11 @@ jobs:
           override: true
     - uses: actions/checkout@v3
     - name: Run benchmark
-      run: cargo bench --bench all_benchmarks -- --output-format bencher | tee output.txt
+      run: |
```
mpaulucci, line 25: this means that if a new bench is added, it has to be added here?

xqft: Yes, if it makes sense to run them on here. For example this job is missing the `criterion_metal` benchs that should execute on a different job running on our M2s, and `criterion_field` which benchs nanosecond ops between field elements, whose results get totally drowned with the CI's generated noise (check out [here](https://lambdaclass.github.io/lambdaworks/bench/)).

mpaulucci: makes sense, thx!

mpaulucci (approved):

gpu/benches/functions/metal.rs
Oppen, line 1: I'm not sure IAI is a good fit here. I'm assuming this runs mostly in the GPU, right? This means emulating main memory and CPU cache behavior will be a bad predictor of wall clock time.

xqft: You're right! this won't work. I'll leave only criterion benches. Thanks!

xqft: Oooh wait I think I never added them in the first place! Because there's no way to compile the `metal` module in the CI yet.

mpaulucci: If you think with main you should be able to get meta in the ci

xqft: Solved in latest commits

xqft:
Closed in favour of #268, will cherry-pick some things from here

.github/workflows/iai_benchs_main.yml
```diff
@@ -0,0 +1,33 @@
...
+    - name: Install valgrind
+      run: |
+        sudo apt update
+        sudo apt-get install -y valgrind
+    - name: Install stable toolchain
+      uses: dtolnay/rust-toolchain@stable
+      with:
+          toolchain: stable
+    - uses: actions/checkout@v3
+    - name: Initialize iai cache for ${{ github.event.pull_request.base.sha }}
```
Oppen, line 21: This was a mistake on my part in the original pipeline. It worked pretty much by miracle. The correct line is:
```suggestion
    - name: Initialize iai cache for ${{ github.sha }}
```
That's because for push to main there's not actual pull request event, and the valid source of your commit hash is the one from the base context.

Oppen: Oh, it worked because it's just the message lol

Oppen (approved):

pablodeymo (approved):
LGTM

PR #258, opened 2023-04-14
Use macOS self hosted runners

klaus993:
These tags make any workflow run in self hosted macOS runners. They have macOS 13 (Ventura), with M2 PRO CPU/GPU

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -53,7 +53,7 @@ jobs:
         run: make clippy
   test:
     name: Test
-    runs-on: macos-latest
+    runs-on: [self-hosted, macOS]
     env:
       CARGO_TERM_COLOR: always
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -72,3 +72,18 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          components: clippy
+      - name: Install cargo-llvm-cov
+        uses: taiki-e/install-action@cargo-llvm-cov
+      - name: Run tests
+        run: cargo test -F metal
```
xqft, line 89:
```suggestion
        run: cargo test -F metal -- metal
```
for executing only those in the `metal` module

mpaulucci: Im not sure about that, I would like to run all test that use metal indirectly too

xqft: you're right! my bad

xqft (approved):

PR #265, opened 2023-04-17
Compilation in ESP32

pablodeymo:
# Compilation in ESP32

## Description

Scripts and crates for the compilation for the target ESP32 model C3 (with Risc V architecture).

Installation pre-requirements:
```
rustup target add riscv32imc-unknown-none-elf
cargo install cargo-generate
cargo install ldproxy
cargo install espup
cargo install espflash
```

(source: https://hutscape.com/tutorials/blinky-rust-esp32c3#install-dependancies)

to compile for this target, in the folder `proving_system/stark` run:
`make compile_esp`

## Type of change

- [ ] New feature
- [ ] Bug fix
- [ ] Optimization

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,4 +9,5 @@ members = [
     "proving_system/stark",
     "proving_system/plonk",
     "gpu",
+    "test_esp32",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,6 @@
+[profile.release]
+opt-level = "s"
+
 [workspace]
 members = [
     "math",
--- a/crypto/Cargo.toml
+++ b/crypto/Cargo.toml
@@ -7,6 +7,7 @@ edition = "2021"
 
 [dependencies]
 sha3 = "0.10.6"
+sha2 = { version = "0.10.6", optional = true }
 embedded-hal = { version = "1.0.0-alpha.9", optional = true }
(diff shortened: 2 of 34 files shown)


Cargo.toml
```diff
@@ -1,3 +1,6 @@
+[profile.release]
+opt-level = "s"
```
Oppen, line 2: I suggest we use a different profile for ESP32. As it is, this change would pessimize runtime for everyone else.

Oppen: I think it can also be done in the same profile as long as the target is specified like with other options.

crypto/Cargo.toml
```diff
@@ -5,14 +5,22 @@ edition = "2021"
...
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
-[dependencies.lambdaworks-math]
-path = "../math"
-
 [dependencies]
 sha3 = "0.10.6"
+embedded-hal = { version = "1.0.0-alpha.9", optional = true }
+esp-idf-sys = { version = "0.32", features = ["binstart"], optional = true }
+esp-idf-hal = { version = "0.40.1", optional = true }
```
Oppen, line 12: Similarly, here we should make these dependencies target dependent.

Oppen (commented):
Nice!

PR #267, opened 2023-04-20
Hash to field elements function

IAvecilla:
# Hash to field elements function

## Description

We need a function that takes a byte message, hash it, and then produces an output of _n_ elements of a finite field. All this algorithm is based on the explanation developed in this [paper](https://www.ietf.org/id/draft-irtf-cfrg-hash-to-curve-16.html#name-hashing-to-a-finite-field).
For now, we only have a test to check that there are no runtime errors. We should add a test to check that the generated fields follow a uniform distribution (See #273).

## Type of change

- [x] New feature

## Checklist
- [x] Linked to Github Issue
- [x] Unit tests added


crypto/src/hash/hash_to_field.rs
```diff
@@ -0,0 +1,91 @@
...
+
+use lambdaworks_math::{
+    field::{
+        element::FieldElement,
+        fields::montgomery_backed_prime_fields::{IsModulus, MontgomeryBackendPrimeField},
+    },
+    unsigned_integer::element::UnsignedInteger,
+};
+
+pub fn hash_to_field<M: IsModulus<UnsignedInteger<N>> + Clone + Debug, const N: usize>(
```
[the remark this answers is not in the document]
gabrielbosio: They are needed to implement `IsField`: https://github.com/lambdaclass/lambdaworks/blob/main/math/src/field/fields/montgomery_backed_prime_fields.rs#L89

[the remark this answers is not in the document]
MauroToscano: Let's make another issue for this!

MauroToscano (approved):

gabrielbosio (approved):

crypto/src/hash/hash_to_field.rs
```diff
@@ -0,0 +1,92 @@
...
+    let k = (log2_p / 2.0).ceil() * 8.0;
+    let l = (((log2_p * 8.0) + k) / 8.0).ceil() as usize;
+    for (i, item) in u.iter_mut().enumerate() {
+        let elm_offset = l * i;
+        let tv = &pseudo_random_bytes[elm_offset..elm_offset + l];
+
+        *item = os2ip::<M, N>(tv);
+    }
+    u
+}
```
schouhy, line 29: I think we can replace floating point operations and replace them with shifts. That way we can also extract the computation of `l` to a const function, since it doesn't depend on any other parameter at the moment. Here's a snippet of what I mean. But written this way seems a bit weird to define `k` altogether. I would omit it or take it as a paramter instead. What do you think?
```suggestion
const fn compute_length<const N: usize>(order: &UnsignedInteger<N>) -> usize {
    let log2_p = order.limbs.len() << 3;
    let k = log2_p >> 2;
    ((log2_p << 3) + k) >> 3
}

/// Converts a hash, represented by an array of bytes, into a vector of field elements
/// For more info, see
/// https://www.ietf.org/id/draft-irtf-cfrg-hash-to-curve-16.html#name-hashing-to-a-finite-field
pub fn hash_to_field<M: IsModulus<UnsignedInteger<N>> + Clone, const N: usize>(
    pseudo_random_bytes: &[u8],
    count: usize,
) -> Vec<FieldElement<MontgomeryBackendPrimeField<M, N>>> {
    let order = M::MODULUS;
    let mut u = vec![FieldElement::zero(); count];
    //L = ceil((ceil(log2(p)) + k) / 8), where k is the security parameter of the cryptosystem (e.g. k = ceil(log2(p) / 2))
    let l = compute_length::<N>(&order);
    for (i, item) in u.iter_mut().enumerate() {
        let elm_offset = l * i;
        let tv = &pseudo_random_bytes[elm_offset..elm_offset + l];

        *item = os2ip::<M, N>(tv);
    }
    u
}
```

gabrielbosio: You're right, we can avoid using floats and extract the length calculation into a different function, omitting `k`. I will do that.

gabrielbosio: I think `k` should be `log2_p << 2` instead of `log2_p >> 2`

schouhy (approved):

PR #268, opened 2023-04-21
Implemented parallel FFT on CUDA

xqft:
# Implemented parallel FFT on CUDA

## Description

Implemented parallel FFT con CUDA and added support for polynomial evaluation and interpolation.

## Type of change

- New feature


gpu/src/cuda/fft/ops.rs
```diff
@@ -0,0 +1,116 @@
...
+        #[test]
+        fn test_cuda_fft_matches_sequential_fft(input in field_vec(2)) {
+            let order = input.len().trailing_zeros();
+            let twiddles = get_twiddles(order.into(), RootsConfig::BitReverse).unwrap();
+
+            let mut cuda_fft = super::fft(&input, &twiddles);
+            lambdaworks_fft::bit_reversing::in_place_bit_reverse_permute(&mut cuda_fft);
+            let fft = lambdaworks_fft::ops::fft(&input).unwrap();
+
+            assert_eq!(cuda_fft, fft);
```
mpaulucci, line 113: prop_assert_eq!?

gpu/src/cuda/field/element.rs
```diff
@@ -0,0 +1,42 @@
...
+    }
+}
+
+impl<F: IsField> Default for CUDAFieldElement<F> {
+    fn default() -> Self {
+        Self { value: F::zero() }
+    }
+}
+
+unsafe impl<F: IsField> DeviceRepr for CUDAFieldElement<F> {
```
mpaulucci, line 24: couldn't you implement this trait for FFTField instead and not create this abstraction?

gpu/src/cuda/shaders/fp.cuh
```diff
@@ -0,0 +1,186 @@
+// Copyright 2023 RISC Zero, Inc.
```
mpaulucci, line 1: can we remove this file?

xqft: yep

mpaulucci (approved):

MauroToscano (approved):

gpu/src/cuda/shaders/u128.cuh
```diff
@@ -0,0 +1,203 @@
...
+
+    // // remove carry from current digit
+    // // fourth32 &= 0xffffffff; // TODO: figure out if this is a nop
+    // // third32 &= 0xffffffff;
+    // second32 &= 0xffffffff;
+    // // first32 &= 0xffffffff;
+
+    // // combine components
+    // // return u128((first32 << 32) | second32, (third32 << 32) | fourth32);
+    // return u128((first32 << 32) | second32, (third32 << 32) | fourth32);
```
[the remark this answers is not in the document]
MegaRedHand: Removed on [`b9f641c`](https://github.com/lambdaclass/lambdaworks/pull/268/commits/b9f641cd2c11c27036115a734959aa7f7e1d6ae0)

fft/benches/criterion_fft.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(dead_code)] // clippy has false positive in benchmarks
```
Oppen, line 1: It's not really false positives but a config where these lines actually don't get built: https://github.com/lambdaclass/lambdaworks/pull/268/files#diff-bd200755e9353b007a0dc108f89d6f18a344836104054d74793c37669718d24cR74-R104

I think it's OK to fix it like this, just pointing out it's not a bug or anything.

MegaRedHand: I removed it and clippy doesn't show any warnings 🤔

gpu/src/cuda/shaders/fp_u256.cuh
```diff
@@ -0,0 +1,210 @@
...
+      exp >>= 1;
+      *this = *this * *this;
+    }
+
+    return res;
+  }
+
+  __device__ Fp256 inverse() {
+    // used addchain
+    // https://github.com/mmcloughlin/addchain
```
Oppen, line 54: Neat trick! I'll probably steal it for cairo 👀

gpu/src/cuda/shaders/u128.cuh
```diff
@@ -0,0 +1,135 @@
...
+  __device__ u128 operator*=(const u128 rhs) {
+    *this = *this * rhs;
+    return *this;
+  }
+
+  unsigned long high;
+  unsigned long low;
+
+  // TODO: Could get better performance with  smaller limb size
+  // Not sure what word size is for M1 GPU
```
Oppen, line 125: Is this relevant with CUDA? Doesn't it only support Metal?

gpu/src/cuda/shaders/u256.cuh
```diff
@@ -0,0 +1,152 @@
...
+  __device__ constexpr bool operator>=(const u256 rhs) const {
+    return !(*this < rhs);
+  }
+
+  __device__ constexpr bool operator<=(const u256 rhs) const {
+    return !(*this > rhs);
+  }
+
+  __device__ constexpr inline u256 operator>>(unsigned shift) const {
+    // TODO: reduce branch conditions
```
Oppen, line 67: That can be done as a sum of products with the bools I think.

gpu/src/cuda/shaders/fp_u256.cuh
MegaRedHand: Be my guest!

gpu/src/cuda/shaders/u256.cuh
Oppen: Something like:
```cpp
u128 new_low = (shift == 0) * low | (shift == 128) * high | (shift < 128 ^ shift == 0) * (high << (128 - shift) | (low >> shift)) | (256 > shift & shift > 128) * (high >> (shift - 128));
u128 new_high = (shift == 0) * high | (shift < 128 ^ shift == 0) * (high >> shift);
return u256(new_high, new_low);
```

gpu/src/cuda/shaders/u256.cuh
```diff
@@ -0,0 +1,152 @@
...
+
+  __device__ u256 operator*(const bool rhs) const {
+    return u256(high * rhs, low * rhs);
+  }
+
+  __device__ u256 operator*(const u256 rhs) const {
+    // split values into 4 64-bit parts
+    u128 top[2] = {u128(low.high), u128(low.low)};
+    u128 bottom[3] = {u128(rhs.high.low), u128(rhs.low.high),
+                      u128(rhs.low.low)};
```
Oppen, line 96: I'm confused, I see 5 64-bit parts here while the comment says 4.

gpu/src/cuda/shaders/u128.cuh
MegaRedHand: No, most code was ported from the metal implementation and this comment wasn't updated. Changed for a reminder about a possible optimization in changing the limb size.

gpu/src/cuda/shaders/u256.cuh
```diff
@@ -0,0 +1,152 @@
+// https://github.com/andrewmilson/ministark/blob/main/gpu-poly/src/metal/u256.h.metal
```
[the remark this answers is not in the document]
MegaRedHand: Fixed. Also changed to a permalink

gpu/src/cuda/shaders/u256.cuh
MegaRedHand: Thanks for the hack! I added it, but there are some parts of the code that could use a similar approach. As it's an optimization, I'm gonna leave that change for future PRs.

gpu/src/cuda/fft/polynomial.rs
```diff
@@ -0,0 +1,84 @@
...
+// TODO: remove when fft works on non-multiple-of-two input length
+fn log2(n: usize) -> Result<u64, CudaError> {
+    if !n.is_power_of_two() {
+        return Err(CudaError::InvalidOrder(n));
+    }
+    Ok(n.trailing_zeros() as u64)
+}
+
+#[cfg(test)]
+mod tests {}
```
Oppen, line 84: This is curious. Is this being tested elsewhere? If so, it's OK to just remove this.

gpu/src/cuda/shaders/u128.cuh
```diff
@@ -0,0 +1,167 @@
...
+  {
+    return u128(high * rhs, low * rhs);
+  }
+
+  __device__ u128 operator*(const u128 rhs) const
+  {
+    unsigned long t_low_high = __umul64hi(low, rhs.high);
+    unsigned long t_high = __umul64hi(low, rhs.low);
+    unsigned long t_high_low = __umul64hi(high, rhs.low);
+    unsigned long t_low = low * rhs.low;
```
Oppen, line 152: I'm not 100% sure I understand the `t_low_high` and  `to_high_low` computation.
I would guess `__umul64hi` returns the high bits of that product. But wouldn't we want the low bits for those numbers? What am I missing?

gpu/src/cuda/shaders/u128.cuh
```diff
@@ -0,0 +1,167 @@
...
+    }
+    else if ((128 > shift) && (shift > 64))
+    {
+      return u128((low >> (shift - 64)), 0);
+    }
+    else
+    {
+      return u128(0);
+    }
+  }
```
Oppen, line 134: You can use an analogous trick to that of `u256` here.

gpu/src/cuda/fft/polynomial.rs
MegaRedHand: Yes. We added it because we were thinking of adding tests, but later found that's covered with ones in _fft_ crate (just need to run them with `-F cuda`; same for metal).

PR #269, opened 2023-04-24
Example reading file with cairo-rs

pablodeymo:
# Example reading file with cairo-rs

## Description

Implementation of a new crate integrating parsing with cairo-rs and then parsing with lambdaworks.

## Type of change

- [ ] New feature
- [ ] Bug fix
- [ ] Optimization

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run


proving_system/integration_cairo/src/main.rs
```diff
@@ -0,0 +1,219 @@
...
+        // col 32
+        vec![FE::from(0x1b), FE::from(0x1b), FE::from(0x51)],
+        // col 33 - Selector column
+        // vec![FE::one(), FE::one(), FE::zero()],
+    ]);
+
+    assert_eq!(execution_trace.cols(), expected_trace.cols());
+}
+
+fn run(
```
entropidelic, line 135: I think we may want to export this function and use it from the starks crate

proving_system/integration_cairo/src/main.rs
```diff
@@ -0,0 +1,219 @@
...
+    entrypoint: &str,
```
entropidelic, line 136: This argument will be almost always the "main" string, I think we could make it an option and set "main" as the default value when it is None

proving_system/integration_cairo/src/main.rs
```diff
@@ -0,0 +1,219 @@
...
+    Runner(#[from] CairoRunError),
+    #[error(transparent)]
+    EncodeTrace(#[from] EncodeTraceError),
+    #[error(transparent)]
+    VirtualMachine(#[from] VirtualMachineError),
+    #[error(transparent)]
+    Trace(#[from] TraceError),
+}
+
+fn main() {
```
entropidelic, line 35: This seems a little better as a test rather than a main function, what do you  think?

proving_system/stark/src/cairo_run/parser.rs
```diff
@@ -0,0 +1,187 @@
...
+    Runner(#[from] CairoRunError),
+    #[error(transparent)]
+    EncodeTrace(#[from] EncodeTraceError),
+    #[error(transparent)]
+    VirtualMachine(#[from] VirtualMachineError),
+    #[error(transparent)]
+    Trace(#[from] TraceError),
+}
+
+pub fn parse_cairo_file(
```
entropidelic, line 24: Rather than parsing, a better name could be something like `run_program`, and a little documentation explaining that runs a compiled cairo program with cairo-rs and it writes two files into disk, one for the memory and one for the trace would be nice

entropidelic (approved):
LGTM!

jrchatruc (approved):

PR #275, opened 2023-04-25
Added support to all FFT implementations for inputs of all sizes via zero padding

### LINKED ISSUE #274: Add support to all FFT implementations for inputs of all sizes via zero padding
Currently there's no way of doing FFT with inputs whose sizes are not powers of two. A solution for this is to fill the input with `0` elements until a power of two size is reached.

Also handle the case where an input is empty or null (null being that `input = [0, 0, ..., 0]`) to improve the robustness of the FFT library.

xqft:
# Added support to all FFT implementations for inputs of all sizes via zero padding

## Description

Closes #274 

## Type of change

- New feature

## Checklist
- [x] Linked to Github Issue


fft/src/helpers.rs
```diff
@@ -1,10 +1,16 @@
-use super::errors::FFTError;
+use lambdaworks_math::field::{element::FieldElement, traits::IsField};
 
-pub fn log2(n: usize) -> Result<u64, FFTError> {
-    if !n.is_power_of_two() {
-        return Err(FFTError::InvalidOrder(
-            "The order of polynomial + 1 should a be power of 2".to_string(),
-        ));
+pub fn next_power_of_two(n: usize) -> usize {
```
mpaulucci, line 3: please add two unit tests, one for an input that is a power of two and one for a number that isn't.

fft/src/ops.rs
```diff
@@ -7,22 +7,28 @@ use crate::roots_of_unity::get_twiddles;
...
+    fft_iterative::in_place_nr_2radix_fft,
 };
 
+use super::helpers;
+
 /// Executes Fast Fourier Transform over elements of a two-adic finite field `F` in a coset. Usually used for
 /// fast polynomial evaluation.
 pub fn fft_with_blowup<F: IsTwoAdicField>(
-    coeffs: &[FieldElement<F>],
+    input: &[FieldElement<F>],
```
mpaulucci, line 18: Hmm, what was the thought process behind the rename?

mpaulucci (approved):
Great work! +1 after adding unit tests.

fft/src/ops.rs
xqft: I think these functions were first made specifically for polynomial evaluation, and later the API got generalized for any slice of field elements, but these names remained referencing the coefficients of a polynomial. So `input` is a better name (and it matches other `ops.rs` files). This isn't really related to the issue but it was annoying me lol

fft/src/helpers.rs
xqft: Done!

pablodeymo (approved):
LGTM

PR #276, opened 2023-04-26
Fix metal benchmarks.

mpaulucci:


xqft (approved):
Good catch! This would've been detected when #255 merged (the criterion benches jobs only run on main, not on PRs), we may need a job to build all benches on PRs without running them to detect breaking changes.

entropidelic (approved):

pablodeymo (approved):
LGTM

PR #234, opened 2023-04-03
Add FFT and polynomial evaluation/interpolation docs to the project book

xqft:
# Add FFT and polynomial evaluation/interpolation docs to the project book

## Description

This PR introduces various sections to the mdbook explaining theoretical and practical details of the FFT library and its application in polynomial evaluation and interpolation.

## Type of change

- [ ] Documentation

PR #284, opened 2023-04-26
Add support to dealing with multiple fields in metal

mpaulucci:
Until now, we only supported Stark Fields.

With this change, other fields may be supported as well. A file like "fields/stark256.h.metal" has to be created for the new field and the function `field_name` has to be overwritten for it to work.

Additionaly, renamed `IsTwoAdic` trait to the more general `IsFFTField`.


gpu/src/metal/shaders/fields/stark256.h.metal
```diff
@@ -0,0 +1,63 @@
+#pragma once
```
mpaulucci, line 1: centralized all specific instances that use stark field here.

math/src/field/traits.rs
```diff
@@ -21,12 +21,16 @@ pub enum RootsConfig {
...
 /// any other root of unity we need to perform FFT.
-pub trait IsTwoAdicField: IsField {
+pub trait IsFFTField: IsField {
     const TWO_ADICITY: u64;
     const TWO_ADIC_PRIMITVE_ROOT_OF_UNITY: Self::BaseType;
 
+    fn field_name() -> String {
+        "".to_string()
+    }
+
```
xqft, line 31:
```suggestion

    /// Used for searching this field's implementation in other languages, e.g in MSL
    /// for executing parallel operations with the Metal API.
    fn field_name() -> String {
        "".to_string()
    }

```

xqft (approved):
Nice! Also documentation on how to implement a custom field will need to be added in a future PR or maybe in #234.

gabrielbosio (approved):

PR #290, opened 2023-04-28
Sqrt

MauroToscano:
# Sqrt


xqft (approved):
I would add some unit tests for `legendre_symbol()`, other than that LGTM.

MauroToscano:
> I would add some unit tests for `legendre_symbol()`, other than that LGTM.

All three cases of legendre_symbol() are tested through Sqrt

math/src/field/element.rs
```diff
@@ -362,6 +362,81 @@ impl<F: IsPrimeField> FieldElement<F> {
...
+        let (zero, one, two) = (Self::zero(), Self::one(), Self::from(2));
+
+        let mut q = Self::zero() - &one;
+        let mut s = Self::zero();
+
+        while q.is_even() {
+            s = s + &one;
+            q = q / &two;
+            println!("Qi: {:?}", q);
+        }
```
Oppen, line 402: Suggestion: check the trailing zeros if you can (many integer types have a method for that). That's `s` and shifting `s` places to the right gives you `q`.

MauroToscano: It's a good idea, but the implementation is too generic there to make that optimization. I could make this optimization if I define the sqrt only when our UnsignedIntegers are used. Since it's not a critical operation, it's only used for deserealization,  I think it's better to skip this optimization right now

math/src/field/element.rs
```diff
@@ -362,6 +362,81 @@ impl<F: IsPrimeField> FieldElement<F> {
...
+{
+    // Returns the two square roots of `self` if it exists
+    // `None` if it doesn't 
+    pub fn sqrt(&self) -> Option<(Self, Self)> {
+        match self.legendre_symbol() {
+            0 => return Some((Self::zero(), Self::zero())), // self is 0
+            -1 => return None,                              // self is quadratic non-residue
+            1 => (),                                        // self is quadratic residue
+            _ => unreachable!(),
+        };
```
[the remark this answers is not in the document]
MauroToscano: Done

entropidelic (approved):
LGTM!

gabrielbosio (approved):

PR #293, opened 2023-05-02
Remove generics from Stark Prover and Verifier

### LINKED ISSUE #291: Remove generics from Stark Prover and Verifier
Stark Prover and Verifier is generically defined to work with any Field. This creates some issues, since there are algorithms where we can't provide an implementation for any Field. Some cases are:

- Hashes (Poseidon constants are field specific, traditional hash to function depend of the inner implementation of the limbs)
- FFT in GPUs 

In the future we will want to support the Stark252PrimeField field and the mini goldilocks (Ecgfp5), but since right now we just care about Stark252PrimeField we can remove the generic.

To keep it working for two fields in the future, we can use an enum as a configuration paramater, but it's out of the scope of this issue

gabrielbosio:
## Description

Resolves #291.
Removes generics from the hasher methods. This includes everything that uses those methods.
**NOTE:** We should merge #297 first and resolve conflicts.

## Type of change

- Bug fix

## Checklist
- [x] Linked to Github Issue
- [x] Unit tests added
  - [ ] Benchmarks added/run


MauroToscano (approved):

gabrielbosio:
Converting to draft until we merge #267.

MauroToscano:
We will remove generics as needed when changing the protocol to match the Starkware verifier

Issue #259, opened 2023-04-14
Commit trace polynomials before getting OOD element


gabrielbosio:
#233 adds commitment of trace polynomials, but it does it after getting the OOD element `z`. We need to commit the trace polynomials before that.

MauroToscano:
This is done

PR #295, opened 2023-05-03
Synchronize code and protocol

schouhy:
# Synchronize code and protocol

## Description

- This PR refactors the prover and verifier code to sync with the protocol. It incorporates missing verifications and commitments. In paritcular it solves Issue #259.
- Moves computation of the transition zerofiers outside of the main loop of the computation of the composition polynomial. This improves proving time by about a factor of 10.
- Adds support for RAP
- Add tests related to LDE domains (instantiation, evaluation of polynomials).
- Add protocol to the documentation.

- [x] Bug fix


proving_system/stark/tests/integration_tests.rs
```diff
@@ -214,52 +207,83 @@ fn test_prove_cairo_fibonacci() {
...
     assert!(verify(&result, &cairo_air));
 }
 
-#[test_log::test]
-fn test_malicious_trace_does_not_verify() {
-    // A valid execution trace is built from the call_func.cairo binary files.
-    let base_dir = env!("CARGO_MANIFEST_DIR");
-    let dir_trace = base_dir.to_owned() + "/src/cairo_vm/test_data/call_func.trace";
-    let dir_memory = base_dir.to_owned() + "/src/cairo_vm/test_data/call_func.mem";
+// #[test_log::test]
```
entropidelic, line 226: Should this test be removed for now?

proving_system/stark/tests/integration_tests.rs
```diff
@@ -214,52 +207,47 @@ fn test_prove_cairo_fibonacci() {
...
-        coset_offset: 3,
+    let context = AirContext {
+        options: ProofOptions {
+            blowup_factor: 2,
+            fri_number_of_queries: 1,
+            coset_offset: 3,
+        },
+        trace_columns: 3,
+        trace_length: trace_cols[0].len(),
+        transition_degrees: vec![1],
```
entropidelic, line 243: This should be a vector of two elements, `vec![1, 1]`, since we need the transition degrees for each transition. In this case we have the main and the aux transition, both of degree 1.

entropidelic (changes requested):

proving_system/stark/tests/integration_tests.rs
schouhy: good catch!!

entropidelic (approved):
LGTM, amazing work! 🎉

MauroToscano (approved):

mpaulucci (approved):

proving_system/stark/src/fri/mod.rs
```diff
@@ -1,113 +1,117 @@
...
+    (last_value, fri_layer_list)
+}
 
-        last_domain = domain_i.clone();
+pub fn fri_query_phase<F: IsFFTField, A: AIR<Field = F>, T: Transcript>(
+    air: &A,
+    domain: &Domain<F>,
+    fri_layers: &Vec<FriLayer<F>>,
+    transcript: &mut T,
+) -> (Vec<FriDecommitment<F>>, usize)
```
pablodeymo, line 75: what's the return of this function (specially thre second arg)?

pablodeymo (approved):
LGTM!

PR #302, opened 2023-05-04
Rework on Serialization to allow G2 points to be serialized

MauroToscano:
# Serialization rework

## Description

Rework on Serialization to allow G2 points to be serialized. This is a first step to make SRS serialization possible

## Type of change

- [x ] New feature

### DIFF
--- a/math/src/elliptic_curve/short_weierstrass/point.rs
+++ b/math/src/elliptic_curve/short_weierstrass/point.rs
@@ -137,7 +137,10 @@ pub enum PointFormat {
 }
 
 #[derive(PartialEq)]
-pub enum FieldEndianness {
+/// Describes the endianess of the internal types of the types
+/// For example, in a field made with limbs of u64
+/// this is the endianess of those u64
+pub enum Endianness {
     BigEndian,
     LittleEndian,
 }
@@ -148,7 +151,7 @@ where
     FieldElement<E::BaseField>: ByteConversion,
 {
     /// Serialize the points in the given format
-    pub fn serialize(&self, _point_format: PointFormat, endianness: FieldEndianness) -> Vec<u8> {
+    pub fn serialize(&self, _point_format: PointFormat, endianness: Endianness) -> Vec<u8> {
         // TODO: these can be more efficient.
         // More options for point formats should be added
         // E.g: Store the x value, the bit to indicate y.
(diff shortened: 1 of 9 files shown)


math/src/traits.rs
```diff
@@ -24,6 +24,15 @@ pub trait ByteConversion {
         Self: std::marker::Sized;
 }
 
+/// Serialize function without options
+/// Used to serialize data to feed it into Fiat Shamir
+/// in some protocols, when ByteConversion is not avaible
+/// For example, when using Curve Points
+pub trait SimpleSerialization {
+    /// Returns the byte representation of the element in big-endian order.
```
schouhy, line 32: The `big-endian` may not make sense in certain situations.

math/src/traits.rs
```diff
@@ -24,6 +24,15 @@ pub trait ByteConversion {
         Self: std::marker::Sized;
 }
 
+/// Serialize function without options
+/// Used to serialize data to feed it into Fiat Shamir
+/// in some protocols, when ByteConversion is not avaible
+/// For example, when using Curve Points
+pub trait SimpleSerialization {
```
schouhy, line 31: Can we name this `DefaultSerialization` or maybe more specific `FiatShamirSerialization`? Or maybe just `Serialization`? Also I would use an adjective (Serializable) or verb for the trait name.

math/src/elliptic_curve/short_weierstrass/point.rs
```diff
@@ -129,80 +128,102 @@ impl<E: IsShortWeierstrass> IsGroup for ShortWeierstrassProjectivePoint<E> {
...
+#[derive(PartialEq)]
+pub enum PointFormat {
+    Projective,
+    // TO DO:
+    // Uncompressed,
+    // Compressed,
+}
+
+#[derive(PartialEq)]
+pub enum FieldEndianness {
```
schouhy, line 140: Maybe `BaseFieldEndianness`?

schouhy (approved):
Looks good! Left a few minor comments mainly on naming conventions. Feel free to ignore them.

gabrielbosio (approved):

math/src/traits.rs
MauroToscano: It may be used somewhere else so I'd avoid `FiatShamirSerialization`. I wanted to avoid a plain `Serializable` since there are additional serialization functions, and I don't want to call the function `serialize()` to avoid collision of names. I like the default thought and the serializable by they seem hard to mix. 

We could call it`Serializable`, and the function `serialize`, and rename the current `serialize` to `serialize_with_args` or something like that.

math/src/traits.rs
MauroToscano: I forgot to delete that description, it's totally wrong, thanks for pointing that

mpaulucci (approved):
LGTM

PR #311, opened 2023-05-05
Add `CudaState` as a wrapper around cudarc primitives

MegaRedHand:


gpu/src/cuda/abstractions/state.rs
```diff
@@ -0,0 +1,98 @@
...
+
+        // Load PTX libraries
+        // TODO: extract into helper method
+        device
+            .load_ptx(
+                Ptx::from_src(/* Library code */ FFT_PTX),
+                /* Library name */ "fft",
+                /* defined functions */ &["radix2_dit_butterfly"],
+            )
+            .map_err(|err| CudaError::PtxError(err.to_string()))?;
```
matias-gonz, line 31: Will we load all modules (`fft`, `gen_twiddles` and `bitrev_permutation`) at new()?

gpu/src/cuda/fft/polynomial.rs
```diff
@@ -17,10 +17,11 @@ where
     F: IsFFTField,
     F::BaseType: Unpin,
 {
+    let state = CudaState::new().unwrap();
```
matias-gonz, line 20: Should we remove this unwrap()?

gpu/src/cuda/abstractions/state.rs
MegaRedHand: I was thinking of moving the module loading to the corresponding `get_<function>` implementations, but I think we should benchmark it first.

gpu/src/cuda/fft/polynomial.rs
MegaRedHand: Removed! Thanks for finding it!

gpu/src/cuda/abstractions/state.rs
xqft: For now don't think too much about these kinds of optimizations, I think PTX loading is pretty small compared to what FFT takes to run

PR #310, opened 2023-05-05
Add gen_twiddles implementation in CUDA

matias-gonz:

PR #312, opened 2023-05-05
Add Bitrev permutation in CUDA

matias-gonz:


MegaRedHand (approved):

gpu/src/cuda/shaders/bitrev_permutation.cu
```diff
@@ -0,0 +1,15 @@
+#include "./fp_u256.cuh"
+#include "./utils.h"
+
+extern "C"
+{
+    __global__ void bitrev_permutation(
+        const p256::Fp *input,
+        p256::Fp *result
+    ) {
+        unsigned index = blockIdx.x * blockDim.x + threadIdx.x;
```
xqft, line 10:
```suggestion
        unsigned index = threadIdx.x;
```
Same as https://github.com/lambdaclass/lambdaworks/pull/310#discussion_r1187481923

xqft (approved):
Excellent!

mpaulucci (approved):

gpu/src/cuda/abstractions/state.rs
```diff
@@ -244,3 +274,67 @@ impl<F: IsField> CalcTwiddlesFunction<F> {
...
+        let config = LaunchConfig {
+            grid_dim,
+            block_dim,
+            shared_mem_bytes: 0,
+        };
+        unsafe {
+            self.function
+                .clone()
+                .launch(config, (&mut self.input, &self.result))
+        }
```
Oppen, line 325: Why is this unsafe? Could you add a comment explaining both why the call is unsafe and why we're doing the right thing and not making things unsound?

gpu/src/cuda/fft/ops.rs
```diff
@@ -51,23 +50,16 @@ pub fn gen_twiddles<F: IsFFTField>(
...
-        if bit_reversed_index > i {
-            input.swap(i, bit_reversed_index);
-        }
-    }
-}
+pub fn bitrev_permutation<F: IsFFTField>(
+    input: Vec<FieldElement<F>>,
+    state: &CudaState,
+) -> Result<Vec<FieldElement<F>>, CudaError> {
+    let result = input.clone();
```
Oppen, line 57: You could pass the same reference twice instead of cloning here. They are not mutable so it should be allowed.

PR #316, opened 2023-05-08
Pairing returns one when one of the points is zero

ajgara:
# Pairing returns one when one of the points is zero
## Description
When one element is zero, the pairing returns one

## Type of change
- [x] Bug fix


diegokingston (approved):

pablodeymo (approved):
LGTM

PR #320, opened 2023-05-08
Add support for multiple fields in CUDA

MegaRedHand:
# Add support for multiple _fields_ in _CUDA_

## Description

This PR adds support for multiple fields in CUDA, by making the current `radix2_dit_butterfly` kernel a template, that will be wrapped by field-specific implementations like `radix2_dit_butterfly_stark256`. On the rust side all libraries are loaded into the device at `CudaState::new`, and later matched by field name.

## Type of change

- [ ] New feature
- [ ] Bug fix
- [ ] Optimization

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run


xqft (approved):
Nice!

gpu/src/cuda/abstractions/state.rs
```diff
@@ -8,11 +8,12 @@ use cudarc::{
 };
 use lambdaworks_math::field::{
     element::FieldElement,
+    fields::fft_friendly::stark_252_prime_field::Stark252PrimeField,
     traits::{IsFFTField, IsField},
 };
 use std::sync::Arc;
 
-const FFT_PTX: &str = include_str!("../shaders/fft.ptx");
+const STARK256_PTX: &str = include_str!("../shaders/fields/stark256.ptx");
```
mpaulucci, line 16: shouldn't this have a more generic name now that we support multiple fields?

gpu/src/cuda/shaders/fields/stark256.cu
```diff
@@ -0,0 +1,21 @@
...
+    using Fp = Fp256<
+        /* =N **/ /*u256(*/ 576460752303423505, 0, 0, 1 /*)*/,
+        /* =R_SQUARED **/ /*u256(*/ 576413109808302096, 18446744073700081664,
+        5151653887, 18446741271209837569 /*)*/,
+        /* =N_PRIME **/ /*u256(*/ 576460752303423504, 18446744073709551615,
+        18446744073709551615, 18446744073709551615 /*)*/
+        >;
+} // namespace p256
+
+extern "C" __global__ void radix2_dit_butterfly(p256::Fp *input, const p256::Fp *twiddles)
```
mpaulucci, line 18: in metal I think the function has the field name appended.

MegaRedHand: Yes, but in CUDA that makes things more difficult. The `load_ptx` function receives a module name and a list of function names. The function names have a _static_ lifetime bound (i.e. are `&'static str`), so they can't be resolved at runtime. I bypassed it by using the module name (that doesn't have lifetime bounds) to differentiate between each implementation, but in turn I need the function names to be fixed and also to have each implementation in a different PTX.

mpaulucci (approved):

MauroToscano (approved):

Oppen (approved):

PR #326, opened 2023-05-09
Implement MSM on CPU using Pippenger's algorithm

MegaRedHand:
# Implement MSM on CPU using Pippenger's algorithm

## Description

This PR implements Multi-Scalar Multiplication on CPU using Pippenger's algorithm, speeding it up by up to _**10 times**_ (as indicated by benchmarks).

The API cannot be used right now in place of the _naive_ implementation, because it requires scalars to be `UnsignedInteger`s. This could be fixed by adding it conditionally via a trait `IsFixedSize` for `BaseType`s that adds a method or constant with the fixed number of bits of the base type of the field.

## Type of change

- [X] Optimization

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [X] Benchmarks added/run

### DIFF
--- a/math/src/msm/pippenger.rs
+++ b/math/src/msm/pippenger.rs
@@ -1,0 +1,127 @@
+use crate::{
+    cyclic_group::IsGroup,
+    field::{
+        element::FieldElement,
+        fields::montgomery_backed_prime_fields::{IsModulus, MontgomeryBackendPrimeField},
+    },
+    unsigned_integer::element::UnsignedInteger,
+};
+use std::fmt::Debug;
+
+pub fn msm<M, G, const NUM_LIMBS: usize>(
+    cs: &[FieldElement<MontgomeryBackendPrimeField<M, NUM_LIMBS>>],
+    hidings: &[G],
+    window_size: usize,
+) -> G
+where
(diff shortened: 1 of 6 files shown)


xqft (approved):
🚀

mpaulucci (approved):

matias-gonz (approved):

math/src/msm/pippenger.rs
```diff
@@ -0,0 +1,146 @@
...
+///
+/// Assume a group G of order r is given.
+/// Let `hidings = [g_1, ..., g_n]` be a tuple of group points in G and
+/// let `cs = [k_1, ..., k_n]` be a tuple of scalars in the Galois field GF(r).
+///
+/// Then, with additive notation, `msm(cs, hidings)` computes k_1 * g_1 + .... + k_n * g_n.
+///
+/// If `hidings` and `cs` are empty, then `msm` returns the zero element of the group.
+///
+/// Panics if `cs` and `hidings` have different lengths.
```
Oppen, line 13: I would recommend it doesn't panic, but if it does, then the `debug_assert_eq!` should be replaced by a regular `assert_eq!`.

MegaRedHand: Changed it. Now it returns a `Result` that is later unwrapped by callers.

PR #328, opened 2023-05-10
Add kzg tests

MauroToscano:
# Add more tests for kzg functions

## Description

Kzg doesn't have a lot of tests, and batch functions are not tested, this adds tests

## Type of change

- [ x ] New feature

## Checklist
- [ x ] Unit tests added

### DIFF
--- a/crypto/src/commitments/kzg.rs
+++ b/crypto/src/commitments/kzg.rs
@@ -346,7 +346,7 @@ mod tests {
 
         let proof = kzg.open_batch(&x, &[y0.clone()], &[p0], &upsilon);
 
-        assert!(!kzg.verify_batch(&x, &[y0], &[p0_commitment], &proof, upsilon));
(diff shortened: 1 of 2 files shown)


crypto/src/commitments/kzg.rs
```diff
@@ -323,6 +323,84 @@ mod tests {
         assert!(kzg.verify(&x, &y, &p_commitment, &proof));
     }
 
+    #[test]
+    fn poly_9000_constant_should_verify_proof() {
+        let kzg = KZG::new(create_srs());
+        let p = Polynomial::<FrElement>::new(&[FieldElement::from(9000)]);
```
schouhy, line 329: I think we could just write `let p = Polynomial::new(&[FieldElement::from(9000)]);`

Oppen (approved):

entropidelic (approved):

PR #333, opened 2023-05-11
Add checked and unchecked from hex

MauroToscano:
# Add checked and unchecked from hex

## Description

Add checked and unchecked from hex to avoid unwanted panics. Most of the prs is the change of calls from `from` to `from_hex_unchecked`. I've also added tests for the checked versions for `UnsignedIntegers` and `FieldElements`through their montgomery backend.

Functions are splitted since we need const calls, but we can't use results there, so we leave the unchecked for creating values in compile time, and checked for runtime.
 
## Type of change

- [ x ] New feature


gabrielbosio (approved):

pablodeymo (approved):
LGTM

PR #336, opened 2023-05-11
Add parallel Pippenger's on CPU via rayon

MegaRedHand:
# Add parallel Pippenger's on CPU via _rayon_

Depends on #326

## Description

This PR adds a parallel version of Pippenger's algorithm with support for rayon's iterator-based parallelization.

## Type of change

- [X] Optimization

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
- [X] This change requires new documentation.
  - [X] Benchmarks added/run


math/src/msm/pippenger.rs
```diff
@@ -86,10 +94,67 @@ where
...
         .unwrap_or_else(G::neutral_element)
 }
 
+#[cfg(feature = "rayon")]
+// It has the following differences with the sequential one:
+//  1. It uses one vec per thread to store buckets.
+//  2. It reduces all window results via a different method.
+pub fn parallel_msm_with<const NUM_LIMBS: usize, G>(
+    cs: &[UnsignedInteger<NUM_LIMBS>],
+    hidings: &[G],
```
schouhy, line 108: Could we name the second parameter `points` or something like that? In this abstract math context the term "hiding" doesn't really have a meaning.

schouhy (approved):
Awesome! Thanks! Left a minor comment

math/src/msm/pippenger.rs
MegaRedHand: Sure. I was just following the convention on the original `msm` function. I'll change it in all of them then.

MegaRedHand: Done

pablodeymo (approved):
LGTM

Oppen (approved):

PR #337, opened 2023-05-11
Add cairo memory and range check constraints

schouhy:
# Add auxiliary constraints 

## Description
Adds the memory and range check constraints. Also fixes some bugs in the stark prover related to the composition polynomial degree bounds, degrees of the cairo transition constraints and the provers use of FFT to evaluate on LDE.

## Type of change
- [x] New feature
- [x] Bug fix

## Checklist
- [x] Unit tests added


proving_system/stark/src/air/example/cairo.rs
```diff
@@ -86,92 +148,320 @@ pub struct PublicInputs {
...
+    v_aux.splice(public_input_section.., public_input.program.clone());
+
+    (a_aux, v_aux)
+}
+
+fn sort_columns_by_memory_address(adresses: Vec<FE>, values: Vec<FE>) -> (Vec<FE>, Vec<FE>) {
+    let mut tuples: Vec<_> = adresses.into_iter().zip(values).collect();
+    tuples.sort_by(|(x, _), (y, _)| x.representative().cmp(&y.representative()));
+    let (adresses, values): (Vec<_>, Vec<_>) = tuples.into_iter().unzip();
+    (adresses, values)
```
entropidelic, line 221: this could be condensed to just

```rust
tuples.into_iter().unzip()
```

proving_system/stark/src/air/example/cairo.rs
```diff
@@ -86,92 +148,320 @@ pub struct PublicInputs {
     pub fp_init: FE,
     pub pc_final: FE,
     pub ap_final: FE,
-    // pub rc_min: u16, // minimum range check value (0 < rc_min < rc_max < 2^16)
-    // pub rc_max: u16, // maximum range check value
-    // pub mem: (Vec<u64>, Vec<Option<FE>>), // public memory
+    pub range_check_min: Option<u16>, // minimum range check value (0 < rc_min < rc_max < 2^16)
+    pub range_check_max: Option<u16>, // maximum range check value
```
entropidelic, line 152: Can you add a brief comment to explain why are these `Options`?

proving_system/stark/src/air/example/cairo.rs
```diff
@@ -195,34 +488,75 @@ impl AIR for CairoAIR {
...
         BoundaryConstraints::from_constraints(constraints)
     }
 
     fn context(&self) -> AirContext {
         self.context.clone()
     }
+
+    fn composition_poly_degree_bound(&self) -> usize {
+        2 * self.context().trace_length
+    }
```
entropidelic, line 559: This bound is calculated with the max transition degree and the trace length, right?
If this is the case, this could be defined directly in the AIR trait with a default implementation, taking the max of the degrees stored in the AirContext.

On another note, should't this be the trace lenght - 1? Maybe I am missing something

entropidelic (commented):
Excellent work guys 🎉 . Left some nits, but overall looks good!

proving_system/stark/src/air/example/cairo.rs
ajgara: Thanks!

docs/src/proving_systems/starks/protocol.md
```diff
@@ -1,64 +1,101 @@
...
-5. Sort $M_{\text{MemRepl}}$ by the first column in increasing order. The result is a matrix $M_{\text{MemReplSorted}}$ of size $(L+L')\times 2$. Denote its columns by `a'` and `b'`.
-6. Compute the vector $p$ of size $4(L+L')$ with entries 
-$$ p_i = \prod_{j=0}^i\frac{z - (\alpha a_i' + v_i')}{z - (\alpha a_i + v_i)}$$
-5. Reshape the matrix $M_{\text{MemReplSorted}}$ into a $(L+L')\times8$ in row-major. Reshape the vector $p$ into a $(L+L')\times4$ matrix in row-major.
-6. Concatenate these 12 rows. The result is a matrix $M'$ of size $(L+L') \times 12$.
-7. Pad $M'$ with copies of its last row until it has a power of two number of rows. As a result we obtain a matrix $M_{\text{RAP2}}\in\mathbb{F}^{2^n\times 12}$.
+1. Stack the rows of the submatrix of $M_{\text{RAP1}}$ defined by the columns `pc, dst_addr, op0_addr, op1_addr` into a vector `a` of length $2^{n+2}$ (this means that the first entries of `a` are `pc[0], dst_addr[0], op0_addr[0], op1_addr[0], pc[1], dst_addr[1],...`).
```
Oppen, line 57: Tip: Markdown supports numbered lists to be defined only as `1.` for all items. It can then fix the numbers itself. This is useful in case in the future you need to add some item in the middle as it saves you from updating all one by one.

proving_system/stark/src/air/constraints/evaluator.rs
```diff
@@ -102,26 +92,26 @@ impl<'poly, F: IsFFTField, A: AIR + AIR<Field = F>> ConstraintEvaluator<'poly, F
...
                 .map(|(index, (boundary_poly, boundary_zerofier))| {
-                    let quotient_degree = boundary_poly.degree() - boundary_zerofier.degree();
-
                     let (boundary_alpha, boundary_beta) =
                         alpha_and_beta_boundary_coefficients[index].clone();
 
                     (boundary_poly.evaluate(d) / boundary_zerofier.evaluate(d))
-                        * (&boundary_alpha
-                            * d.pow(max_degree_power_of_two - (quotient_degree as u64))
+                        * (&boundary_alpha * d.pow(boundary_term_degree_adjustment)
```
Oppen, line 114: We should be able to extract `d.pow(boundary_term_degree_adjustment)` out of the closure to running exponentiation of the same values `boundary_polys.len()` times.

proving_system/stark/src/air/constraints/evaluator.rs
```diff
@@ -168,9 +156,9 @@ impl<'poly, F: IsFFTField, A: AIR + AIR<Field = F>> ConstraintEvaluator<'poly, F
             .zip(constraint_coeffs)
         {
             let zerofied_eval = eval / div.evaluate(x);
-            let zerofied_degree = trace_degree * transition_degree - div.degree();
-            let result =
-                zerofied_eval * (alpha * x.pow(max_degree - (zerofied_degree as u64)) + beta);
+            let degree_adjustment = air.composition_poly_degree_bound()
+                - (air.context().trace_length * (transition_degree - 1));
```
Oppen, line 160: This power looks like it could be extracted out of the loop as well.

docs/src/proving_systems/starks/protocol.md
schouhy: Nice! Didn't know that

Oppen (approved):

Oppen (approved):

entropidelic (approved):
LGTM, let's merge this when the metal tests are passing

Oppen (approved):

MauroToscano (changes requested):
Needs test that fails only with new constraints

proving_system/stark/tests/integration_tests.rs
```diff
@@ -248,6 +297,83 @@ fn test_prove_rap_fib() {
...
+
+    let result = prove(
+        &(malicious_raw_trace, malicious_memory),
+        &cairo_air,
+        &mut public_input,
+    );
+
+    public_input.program = real_program;
+    assert!(!verify(&result, &cairo_air, &public_input));
 }
```
MauroToscano, line 379: This test has a bad name and isn't documented. We already have a couple of malicious provers, it's not clear in this one what's malicious about it.

MauroToscano (changes requested):

MauroToscano (changes requested):

MauroToscano (approved):

PR #347, opened 2023-05-15
Handle prover errors

### LINKED ISSUE #346: [STARK] Handle prover errors
Right now, no errors are handled in the prover. We need to remove `unwrap`s, `panic`s and unchecked array accesses.

gabrielbosio:
## Description

Resolves #346.

## Type of change

- New feature

## Checklist
- [x] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run

### DIFF
--- a/proving_system/stark/benches/functions/stark.rs
+++ b/proving_system/stark/benches/functions/stark.rs
@@ -35,7 +35,7 @@ pub fn prove_fib(trace_length: usize) {
 
     let fibonacci_air = simple_fibonacci::FibonacciAIR::from(context);
 
-    let result = prove(&trace, &fibonacci_air);
+    let result = prove(&trace, &fibonacci_air).unwrap();
     verify(&result, &fibonacci_air);
 }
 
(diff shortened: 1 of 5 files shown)


Oppen (approved):

proving_system/stark/src/air/frame.rs
```diff
@@ -25,30 +29,36 @@ impl<F: IsFFTField> Frame<F> {
...
         offsets: &[usize],
-    ) -> Self {
+    ) -> Result<Self, AIRError> {
         // Get trace length to apply module with it when getting elements of
         // the frame from the trace.
         let trace_steps = trace.n_rows();
         let data = offsets
             .iter()
             .flat_map(|frame_row_idx| {
+                // It's ok to unwrap here because the row index will always be in bounds
```
MegaRedHand, line 58: I think here we are assuming that `self.data`'s rows will always have the same length (i.e. its length is always divisible by `n_rows`). Is that ok?

MauroToscano:
This is heavily outdated, we can re open it later when we need it

PR #355, opened 2023-05-16
bench: migrate to iai-callgrind

### LINKED ISSUE #307: Iai PR benchmarks don't show the difference with main
Iai benchmarks should show improvements/regressions between main and the to-merge branch.

Also consider changing to [iai-callgrind](https://github.com/Joining7943/iai-callgrind), should be pretty straightforward. @Oppen do you know how does [calliper](https://github.com/osiewicz/calliper/) compare with iai-callgrind?

### LINKED ISSUE #350: Migrate iai use in CI benchmarks to iai-callgrind
For reference see lambdaclass/cairo-rs#1167

Oppen:
Fixes #350 
Fixes #307 

## Checklist
- [x] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,3 +7,6 @@ members = [
     "proving_system/plonk",
     "gpu",
 ]
+
+[workspace.dependencies]
+iai-callgrind = "0.3.1"
--- a/.github/workflows/iai_benchs_main.yml
+++ b/.github/workflows/iai_benchs_main.yml
@@ -13,21 +13,20 @@ jobs:
       run: |
         sudo apt update
         sudo apt-get install -y valgrind
+        cargo install --version 0.3.1 iai-callgrind-runner
     - name: Install stable toolchain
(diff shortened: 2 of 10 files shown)


Cargo.toml
```diff
@@ -7,3 +7,6 @@ members = [
     "proving_system/plonk",
     "gpu",
 ]
+
+[workspace.dependencies]
```
MegaRedHand, line 11: Shouldn't this be something like `[workspace.dev-dependencies]`?

Oppen: No. I would have hoped so, but there's only `dependencies` at the workspace level. But don't worry, those only get imported and built when called, so as long as we only add them to crates in the `dev-dependencies` section they won't end up in the release builds.

MegaRedHand (approved):

pablodeymo (approved):

PR #358, opened 2023-05-19
perf: optimize `IsShortWeierstrass::operate_with`

Oppen:
A few powers with constant exponent of `2` and `3` were converted to their multiplication equivalents. A 30-50% speed improvement was measured in benchmarks.

## Type of change
- [x] Optimization

## Checklist
  - [x] Benchmarks run

### DIFF
--- a/math/src/elliptic_curve/short_weierstrass/point.rs
+++ b/math/src/elliptic_curve/short_weierstrass/point.rs
@@ -97,16 +97,17 @@ impl<E: IsShortWeierstrass> IsGroup for ShortWeierstrassProjectivePoint<E> {
                 if u1 != u2 || *py == FieldElement::zero() {
                     Self::neutral_element()
                 } else {
+                    let eight = FieldElement::from(8);
                     let w = E::a() * pz * pz + FieldElement::from(3) * px * px;
                     let w_square = &w * &w;
                     let s = py * pz;
                     let s_square = &s * &s;
                     let b = px * py * &s;
-                    let h = &w_square - FieldElement::from(8) * &b;
+                    let h = &w_square - &eight * &b;
                     let xp = FieldElement::from(2) * &h * &s;
                     let yp = w * (FieldElement::from(4) * &b - &h)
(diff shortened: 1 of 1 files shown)


math/src/elliptic_curve/short_weierstrass/point.rs
```diff
@@ -97,25 +97,30 @@ impl<E: IsShortWeierstrass> IsGroup for ShortWeierstrassProjectivePoint<E> {
...
+                    let s_square = &s * &s;
                     let b = px * py * &s;
-                    let h = w.pow(2_u16) - FieldElement::from(8) * &b;
+                    let h = &w_square - FieldElement::from(8) * &b;
                     let xp = FieldElement::from(2) * &h * &s;
                     let yp = w * (FieldElement::from(4) * &b - &h)
-                        - FieldElement::from(8) * py.pow(2_u16) * s.pow(2_u16);
-                    let zp = FieldElement::from(8) * s.pow(3_u16);
+                        - FieldElement::from(8) * py * py * &s_square;
+                    let zp = FieldElement::from(8) * &s * &s_square;
```
schouhy, line 109: In the case of the Montgomery backend, instantiating `FieldElement::from(8)` involves a call to `cios`. Did you try extracting the coefficients as well to see if that boosts it a little more?

schouhy (approved):
Nice!!

math/src/elliptic_curve/short_weierstrass/point.rs
Oppen: I tried briefly but, not knowing it would end up calling `cios`, I decided to revert it before trying the performance to avoid adding too much noise. I can add it to this PR and run the benches again.

Oppen: Wow this was an insanely good idea @schouhy! Thanks! It improves `evaluate_slice` by almost 80% __on top of__ the original optimization!

MauroToscano (approved):

MegaRedHand (approved):
🚀

PR #359, opened 2023-05-19
perf: optimize `Polynomial::interpolate`

MegaRedHand:
## Description

This PR adds a simple optimization to `Polynomial::interpolate`, that reduces execution time by ~25%, according to benchmarks.

The optimization consists of precomputing denominators, and then inverting them all at once, using the batch inverse algorithm, along with only calculating half of them, as the other half is equal to their negation.

## Type of change

- [ ] New feature
- [ ] Bug fix
- [x] Optimization

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run


Oppen (approved):

math/src/polynomial.rs
```diff
@@ -35,6 +35,9 @@ impl<F: IsField> Polynomial<FieldElement<F>> {
         Self::new(&[])
     }
 
+    /// Returns a polynomial that interpolates the points with x coordinates and y coordinates given by
+    /// `xs` and `ys`.
+    /// `xs` and `ys` must be the same length, and `xs` values should be unique. If not, panics.
     pub fn interpolate(xs: &[FieldElement<F>], ys: &[FieldElement<F>]) -> Self {
         // TODO: try to use the type system to avoid this assert
```
Oppen: If by chance the size is compile-time constant, we could probably do:
```suggestion
    pub fn interpolate<const N: usize>(xs: &[FieldElement<F>; N], ys: &[FieldElement<F>; N]) -> Self {
        // TODO: try to use the type system to avoid this assert
```

Oppen (approved):

MauroToscano (approved):

math/src/polynomial.rs
MegaRedHand: Sadly, it is not 😞

I was thinking maybe unify them to `&[[FieldElement<F>; 2]]`, `&[(FieldElement<F>, FieldElement<F>)]`, some wrapper struct, or an iterator that returns tuples of x and y values.

math/src/polynomial.rs
```diff
@@ -35,17 +35,51 @@ impl<F: IsField> Polynomial<FieldElement<F>> {
...
+        for (i, xi) in xs.iter().enumerate().skip(1) {
+            indexes.push(idx);
+            for (j, xj) in xs.iter().enumerate() {
+                if i > j {
+                    // TODO: return Result here?
+                    assert_ne!(xi, xj, "xs values must be unique");
+                    denominators.push(xi - xj);
+                    idx += 1;
+                }
+            }
```
Oppen, line 62:
```suggestion
            for (j, xj) in xs.iter().enumerate().take(i-1) {
                // TODO: return Result here?
                assert_ne!(xi, xj, "xs values must be unique");
                denominators.push(xi - xj);
                idx += 1;
            }
```

math/src/polynomial.rs
```diff
@@ -557,8 +591,31 @@ impl<F: IsField> ops::Sub<&Polynomial<FieldElement<F>>> for FieldElement<F> {
...
+    }
+    let mut bi_inv = prod_prefix[count - 1].inv();
+    for i in (1..count).rev() {
+        let ai_inv = &bi_inv * &prod_prefix[i - 1];
+        bi_inv = &bi_inv * &numbers[i];
+        numbers[i] = ai_inv;
+    }
+    numbers[0] = bi_inv;
+}
+
```
Oppen, line 614: Possibly move this to field. It will be useful in other places as well.

MegaRedHand: Moved it to a new module `lambdaworks_math::batch_inverse` (similar to `msm`)

math/src/polynomial.rs
```diff
@@ -52,13 +52,11 @@ impl<F: IsField> Polynomial<FieldElement<F>> {
...
         for (i, xi) in xs.iter().enumerate().skip(1) {
             indexes.push(idx);
-            for (j, xj) in xs.iter().enumerate() {
-                if i > j {
-                    // TODO: return Result here?
-                    assert_ne!(xi, xj, "xs values must be unique");
-                    denominators.push(xi - xj);
-                    idx += 1;
-                }
+            for (_, xj) in xs.iter().enumerate().take(i) {
```
Oppen, line 55:
```suggestion
            for xj in xs.iter().take(i) {
```

schouhy (changes requested):
Cool! I would make `inplace_batch_inverse` part of the `FieldElement` impl. What do you think?

MegaRedHand:
@schouhy Done. Although, in the future, we may need to move the implementation to another module and keep `FieldElement::inplace_batch_inverse` as a wrapper that chooses between different versions of the algorithm (CPU sequential/parallel and GPU 🚀) as in `Polynomial::evaluate_fft`

schouhy (approved):
Awesome! Thanks for considering my request!

PR #363, opened 2023-05-22
Fix Metal memory leaks

xqft:
# Fix Metal memory leaks

## Description

[the metal CI tests](https://github.com/lambdaclass/lambdaworks/actions/runs/5029341479/jobs/9020899105?pr=337) of #337 fail because of a Metal memory bug which causes UB. We think this is because Metal's memory is supposed to be handled by Obj-C/Swift's Automatic Reference Counting (ARC), which doesn't exist in Rust and the crate used to provide the APIs bindings doesn't manage this memory neither. A solution for this is to wrap our code in an [autoreleasepool block](https://stackoverflow.com/questions/14677049/what-is-autoreleasepool), which will drain the memory when it Drops. This block hurts performance by a tiny amount, but according to the previous article it may be the only way to manage the memory.

The bug specifically happens when the command buffer and encoder are created outside of the `autoreleasepool` block.

## Type of change

- Bug fix


fft/src/polynomial.rs
```diff
@@ -403,65 +403,6 @@ mod tests {
...
-                poly in poly(4).prop_filter("Avoid non pows of two", |poly| poly.coeff_len().is_power_of_two())) {
-                objc::rc::autoreleasepool(|| {
-                    let (poly, new_poly) = gen_fft_interpolate_and_evaluate(poly);
-                    prop_assert_eq!(poly, new_poly);
-                    Ok(())
-                }).unwrap()
-            }
-        }
-
-        #[cfg(not(feature = "metal"))]
```
Oppen, line 464: Why are these tests being removed?

matias-gonz (approved):

pablodeymo (approved):
LGTM

schouhy (approved):

PR #364, opened 2023-05-22
Add mixed addition optimization for the miller loop of the BLS 12 381 curve

schouhy:
curve

# TITLE

## Description

This PR adds mixed addition optimizations to the miller loop of the bls 12 381 curve ([source](https://eprint.iacr.org/2013/722.pdf)), along with other minor optimizations.

## Type of change
- [x] Optimization

## Checklist
- [x] Unit tests added

### DIFF
--- a/math/src/elliptic_curve/short_weierstrass/curves/bls12_381/field_extension.rs
+++ b/math/src/elliptic_curve/short_weierstrass/curves/bls12_381/field_extension.rs
@@ -1,15 +1,13 @@
-use crate::unsigned_integer::element::U384;
-use crate::{
-    field::{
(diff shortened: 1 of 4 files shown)


Oppen (commented):
I think right now the main algorithm is more than I can chew. Other than that, the code looks good.
I would be interested in seeing the benchmarks.

math/benches/criterion_elliptic_curve.rs
Oppen, line 1: Can we add similar ones for IAI so we can spot regressions in the CI?

Oppen (approved):

ajgara (approved):

PR #365, opened 2023-05-22
Refactor prover verifier

ajgara:
# Refactor of prover and verifier

## Description
It creates structs for the Prover and Verifier. This simplifies some shared arguments for all functions: the AIR and the domain. It also simplifies a bit the code structure and it follows the same API as Plonk.

## Type of change

- [x] Refactor


MauroToscano:
No longer valid

PR #369, opened 2023-05-23
bench: hide results behind black_box to avoid elimination

Oppen:
Several benchmarks were misreporting 0 cycles due to dead code removal. The calls have been wrapped in `black_box` to avoid this. Many benchmark functions have been marked as no inline as well. There are also a few necessary fixes to the iai workflows.

## Type of change

- [x] Bug fix

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run

### DIFF
--- a/.github/workflows/iai_benchs_main.yml
+++ b/.github/workflows/iai_benchs_main.yml
@@ -28,5 +28,5 @@ jobs:
       uses: actions/cache/save@v3
       with:
         path: |
-          */target
-        key: ${{ runner.os }}-iai-benchmark-cache-${{ github.event.pull_request.base.sha }}
+          **/target/iai
+        key: ${{ runner.os }}-iai-benchmark-cache-${{ github.sha }}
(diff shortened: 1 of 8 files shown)


.github/workflows/iai_benchs_main.yml
```diff
@@ -28,5 +28,5 @@ jobs:
       uses: actions/cache/save@v3
       with:
         path: |
-          */target
-        key: ${{ runner.os }}-iai-benchmark-cache-${{ github.event.pull_request.base.sha }}
+          **/target/iai
+        key: ${{ runner.os }}-iai-benchmark-cache-${{ github.event.sha }}
```
MegaRedHand, line 32: Shouldn't this be `github.sha`?

MegaRedHand (approved):
Good catch!

.github/workflows/iai_benchs_main.yml
Oppen: Yes, it should, thanks!

xqft (approved):

PR #370, opened 2023-05-23
chore: refactor `fft` and some modules of `gpu` into `math`

MegaRedHand:
## Description

This PR moves the _fft_ and _gpu_ crates' contents into the _math_ crate, as they are part of the internal implementation and should not be used by consumers.

To avoid making this PR too long and opaque, I am leaving for later PRs:
- updating the FFT benchmarks (they have some scoping problems, and should be made to work with GPU impls just by activating the feature)
- adding a _build-script_ for shader compilation
- moving the GPU-specific FFT implementations into the `math::fft` module (they are currently in the `math::gpu` module)

Adding to this last point: in the future, we will have various different algorithms implemented in the GPU, and so _I think_ it will be better to have the `math::gpu` module only contain common GPU-related utils (of both rust and shader sides).


Oppen (approved):

xqft:
> I think it will be better to have the math::gpu module only contain common GPU-related utils (of both rust and shader sides)
did just that, I'll continue this in a follow up because this is enough refactoring for one PR

diegokingston (approved):

entropidelic (approved):

PR #371, opened 2023-05-24
Compile shaders in ci

MegaRedHand:
## Description

This PR makes the CI compile shaders before running tests and fail if they differ from the committed artifacts. This should help us avoid having un-updated shaders (and broken sources).


Oppen (approved):

xqft:
Wouldn't it be better to make a `build.rs` file which invokes the make rules? That way every time `cargo build` is called, all needed shaders will compile and you also won't need a new CI job.

MegaRedHand:
It depends on if we want to have the shader compilation toolchains as a dependency or not. If we use a build.rs, crate consumers would need to be able to compile the shaders to use our library. Although we _can_ use an environment variable to allow using pre-compiled shaders.

xqft:
I think we should. Both Metal and CUDA can compile into intermediate languages (AIR and PTX respectively), which then need to be assembled to host-specific instructions with CLI tools. We can: or add the toolchain as a dependency, or compile for every target host possible and have the user choose an option. I think the first option is better and reasonable if the user pretends to use CUDA/Metal in their machine.

I don't think if we even want to compile into intermediate code and let the user do the final step, recently I was testing in a remote machine with an 'old' Ubuntu version which doesn't have an updated version of the CUDA toolchain available and I had to recompile PTX files.

MegaRedHand:
Reviewers don't be scared! The 4506 lines removed are from compiled shaders we no longer have in the repository!

mpaulucci (approved):

gpu/build.rs
```diff
@@ -0,0 +1,92 @@
+#[cfg(feature = "cuda")]
+fn compile_cuda_shaders() {
+    use std::process::Command;
+    use walkdir::WalkDir;
+    const CUDA_SOURCE_DIR: &str = "src/cuda/shaders";
```
Oppen, line 5: Does the build script always run from the `CARGO_MANIFEST_DIR`? Or could it be possible that running from `src` breaks this? If the latter, we could fix it reading `CARGO_MANIFEST_DIR` first.

MegaRedHand: I don't really know, but I changed it to use `CARGO_MANIFEST_DIR` so it's more robust.

Oppen (approved):

matias-gonz (approved):

xqft:
-4150, I love this pr

PR #382, opened 2023-05-25
Reorganize `math` folders

xqft:
# Reorganize `math` folders

## Description

This is a follow up of #370, the aim is to have the `math` crate organized like:

```
└── math/
    ├── benches/
    └── src/
        ├── fft/
        │   ├── cpu/
        │   └── gpu/
        │       ├── metal/
        │       └── cuda/
        ├── msm/
        │   ├── cpu/
        │   └── gpu/
        │       ├── metal/
        │       └── cuda/
        └── gpu/
            ├── metal/
            └── cuda/
```
where the `gpu/` folder will contain all the math related shared code needed for the Rust operations in `fft/` and `msm/` (meaning shaders and Fp classes)

## Type of change

- Refactor

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run


gpu/src/cuda/abstractions/state.rs
```diff
@@ -5,8 +5,7 @@ use cudarc::{
 };
 use std::sync::Arc;
 
-const STARK256_PTX: &str =
-    include_str!("../../../../math/src/gpu/cuda/shaders/fields/stark256.ptx");
+const STARK256_PTX: &str = include_str!("../../../../math/src/gpu/cuda/shaders/field/stark256.ptx");
```
Oppen, line 8: Maybe use `CARGO_MANIFEST_DIR` rather than count `../`?

Oppen (approved):

gpu/src/cuda/abstractions/state.rs
xqft: I knew there was a way to get that path! I couldn't find it, thank you!

xqft: Hmm I'm having problems doing this, because `include_str!()` will only take string literals, I tried to use `concatcp!()` from `const_format` but it doesn't like that neither.

mpaulucci (approved):

diegokingston (approved):

PR #386, opened 2023-05-26
Add Metal pippenger benchmarks and improve the bench utils

xqft:
# Add Metal pippenger benchmarks and improve the bench utils

## Description

This PR adds MSM benchmarks for the Metal parallel pippenger algorithm so every improvement can be measured and compared with other implementations.

Also implements standard distribution sampling for `UnsignedInteger`, `FieldElement` and `ShortWeierstrassProjectivePoint<BLS12381Curve>` via the `rand` crate so it's easier to add benchmarks.

## Type of change

- New feature


MauroToscano:
Random maybe troublesome. And this is tale. We can re open it later, but it seems useful

PR #388, opened 2023-05-27
Montgomery inversion on Fp

gabrielbosio:
# TITLE

## Description

## Type of change

- [ ] New feature
- [ ] Bug fix
- [ ] Optimization

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run

### DIFF
--- a/math/src/field/fields/montgomery_backed_prime_fields.rs
+++ b/math/src/field/fields/montgomery_backed_prime_fields.rs
@@ -135,6 +135,7 @@ where
         let mut k = 0;
         let n = M::MODULUS.num_bits();
 
+        // Change shl with x = 2x
         while v > &Self::zero() {
             if u.limbs[0] & 1 == 0 {
                 u.div2();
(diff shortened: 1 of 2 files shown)


math/src/field/fields/montgomery_backed_prime_fields.rs
```diff
@@ -128,7 +128,46 @@ where
...
         }
-        Self::pow(a, M::MODULUS - Self::BaseType::from_u64(2))
+        let mut u = M::MODULUS;
+        let mut v = a.clone();
+        let mut r = Self::zero();
+        let mut s = Self::one();
+        let mut k = 0;
+        let n = M::MODULUS.num_bits();
+
+        while v > Self::zero() {
```
Oppen, line 138: It's __typically__ faster to compare for equality as it should be bit comparison only, while in Montgomery it's not necessarily trivial to compute inequalities (as the comment below implies). Since `v` is unsigned, the only case where `v > 0` is false is when `v == zero`.
```suggestion
        while v != Self::zero() {
```

MauroToscano:
Stale

PR #398, opened 2023-05-31
fix: substraction overflow in `sqrt`

Oppen:
An edge case where the inner loop exits without finding a power such that it becomes `one` makes `i == m` which in turn causes a later `m - i - 1` to overflow.
This was caused by the demotion to primitive type in a previous optimization.
Fixed by promoting back to `FieldElement`.

## Type of change

- [x] Bug fix

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run

### DIFF
--- a/math/src/field/element.rs
+++ b/math/src/field/element.rs
@@ -441,20 +441,20 @@ impl<F: IsPrimeField> FieldElement<F> {
 
         let mut x = self.pow(((&q + &one) / &two).representative());
         let mut t = self.pow(q.representative());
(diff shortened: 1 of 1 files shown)


schouhy (approved):

ajgara (commented):
Can you think of a small test for this edge case? Something that didn't work before this. It looks good to me.

diegokingston (approved):

PR #406, opened 2023-06-02
Implement Square Separated Operand Scanning

schouhy:
# Optimize square

## Description
This PR adds an implementation of the SOS square algorithm ( 2.3.1 [here](https://www.microsoft.com/en-us/research/wp-content/uploads/1998/06/97Acar.pdf) ) and incorporates it to the square and multiply algorithm `pow`.

```
Stark FP operations/sos_square                            
                        time:   [9.8666 ns 9.8868 ns 9.9065 ns]
Stark FP operations/square with mul                       
                        time:   [12.203 ns 12.206 ns 12.209 ns]
```

- [x] Optimization

### DIFF
--- a/math/benches/criterion_field.rs
+++ b/math/benches/criterion_field.rs
@@ -1,7 +1,16 @@
-use criterion::{criterion_group, criterion_main, Criterion, black_box};
-use lambdaworks_math::{field::{
-    element::FieldElement, fields::{fft_friendly::stark_252_prime_field::{Stark252PrimeField, MontgomeryConfigStark252PrimeField}, montgomery_backed_prime_fields::IsModulus},
-}, unsigned_integer::{montgomery::MontgomeryAlgorithms, element::U256}};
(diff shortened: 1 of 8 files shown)


MauroToscano (approved):

math/benches/criterion_field.rs
```diff
@@ -25,6 +36,32 @@ pub fn starkfield_ops_benchmarks(c: &mut Criterion) {
         bench.iter(|| x * y);
     });
 
+    group.bench_with_input("pow by 1", &x.clone(), |bench, x| {
+        bench.iter(|| x.pow(1_u64));
+    });
```
Oppen, line 41: For this and the `square with pow` cases I would add a comment clarifying the non-boxed constant is intentional as it is how it would be used in the wild for their particular use cases.

Oppen (approved):

diegokingston (approved):

PR #409, opened 2023-06-02
Add installation instructions for metal.

mpaulucci:


README.md
```diff
@@ -104,6 +104,22 @@ If you use ```Lambdaworks``` libraries in your research projects, please cite th
 
 ### Gadgets
 
+## Installation
+
+### Metal
+To utilize the GPU on a Mac, it's necessary to install Xcode.
```
MegaRedHand, line 110: Xcode is only necessary to compile the metal shaders. Lib consumers don't need to install it as we already have them committed to the repository.

PR #412, opened 2023-06-04
perf: improve accuracy of field IAI benchmarks

Oppen:
- Extract FP element sample creation into a `util` function to allow
  exclusion from measurement.
- Likewise for getting a quadratic residue by squaring the sample.
- Replace the appropriate call in IAI and Criterion benches.

## Type of change

- [x] Bug fix


pablodeymo (approved):
LGTM

PR #425, opened 2023-06-09
Add bit len to prime fields

MauroToscano:
# Add bit len to prime fields

## Description

Add a function to calculate the amount of bits a prime fields has in their representation. This is needed to update the transcript hash to field function

## Type of change

- [x] New feature

### DIFF
--- a/math/src/field/fields/montgomery_backed_prime_fields.rs
+++ b/math/src/field/fields/montgomery_backed_prime_fields.rs
@@ -257,6 +257,18 @@ where
     fn representative(x: &Self::BaseType) -> Self::RepresentativeType {
         MontgomeryAlgorithms::cios(x, &UnsignedInteger::from_u64(1), &M::MODULUS, &Self::MU)
     }
+
+    fn field_bit_size() -> usize {
+        let mut evaluated_bit = NUM_LIMBS * 64 - 1;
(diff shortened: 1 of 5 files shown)


entropidelic (approved):

diegokingston (approved):

math/src/field/fields/montgomery_backed_prime_fields.rs
```diff
@@ -257,6 +257,18 @@ where
...
+    fn field_bit_size() -> usize {
+        let mut evaluated_bit = NUM_LIMBS * 64 - 1;
+        let max_element = M::MODULUS - UnsignedInteger::<NUM_LIMBS>::from_u128(1);
+        let one = UnsignedInteger::from_u128(1);
+
+        while ((max_element >> evaluated_bit) & one) != one {
+            evaluated_bit -= 1;
+        }
+
+        evaluated_bit + 1
```
Oppen, line 270: If you have access to limbs you can do the [checked_ilog2](https://doc.rust-lang.org/std/primitive.u64.html#method.checked_ilog2) method and operate on the result.

MauroToscano: That seems useful

ajgara (approved):

Oppen (approved):

PR #437, opened 2023-06-16
Add fft benchmark vs winterfell

ajgara:
# Add fft benchmark vs winterfell

## Description

Add a benchmark comparing the FFT implementation of both libraries. It also removes the requirement of the Hash trait for the BaseType of the IsField trait to facilitate the integration of winterfell fields with lambdaworks.


MauroToscano:
We should re add this later, but this PR is no longer valid

PR #438, opened 2023-06-16
Remove Hash trait bound from IsField::BaseType

schouhy:
# Remove Hash trait bound from IsField::BaseType

## Description

Remove unnecessary trait bound


MauroToscano (approved):

diegokingston (approved):

Oppen:
Does this mean users can't create a `HashSet` of field elements or a `HashMap` with field element keys, or is it just that we move that decision to the implementations?

pablodeymo (approved):
LGTM

entropidelic (approved):

PR #442, opened 2023-06-19
use Keccak256 and little endian output in default Transcript

feltroidprime:
Rationale : 
Keccak256 has the same security as Sha3-256 and SHA-3 is not available in Cairo.

Starknet Keccak256 outputs the hash in little endian. 
Using the same representation in the transcript will save many (several hundreds) of Bitshift operations for a Cairo Verifier.


MauroToscano:
We need to fix the tests before merging

MauroToscano (approved):

entropidelic (approved):

schouhy (approved):

PR #443, opened 2023-06-22
No std ff

MauroToscano:
# No STD for Lambdaworks Finite Fields

## Description

No STD for Lambdaworks Fields

Summary:

- Add crate ensure std (Taken from cairo-rs) to make sure there isn't any dependency that requires std
- Updated CI to compile properly, and check `no-std`and the added crate
- Removed `thiserror`from math, since it wasn't compatible with `no-std`
- Disabled `batch_inverse_inplace` since it wasn't std compatible

Notice no std supports only FromBytes implementations of `ByteConversion`for limitation of not having const operations to fix the size of the to_bytes outputs with const generics

For Cairo we have implemented a to_bytes for the specific felt
## Type of change

- [x] New feature


math/src/lib.rs
```diff
@@ -1,10 +1,17 @@
...
 pub mod unsigned_integer;
+// These modules don't work in no-std mode
+#[cfg(not(feature = "no_std"))]
+pub mod elliptic_curve;
+#[cfg(not(feature = "no_std"))]
+pub mod fft;
+#[cfg(not(feature = "no_std"))]
+pub mod msm;
+#[cfg(not(feature = "no_std"))]
+pub mod polynomial;
```
MegaRedHand, line 17: _Note for future work_
If we want to support wasm in the future, we could use a structure similar to [cairo-rs](https://github.com/lambdaclass/cairo-rs/blob/d33220728306855ef2c97bc115b73ebc8ae6c45a/vm/src/lib.rs#L13-L16), with an _stdlib_ module, that uses std or no-std alternatives according to the `no_std` feature.

MegaRedHand (changes requested):
LGTM, but we should use an _std_ feature instead of a _no_std_ one, as this would otherwise cause weird interactions when importing with different features.

math/src/lib.rs
MauroToscano: Done

math/Cargo.toml
```diff
@@ -18,8 +18,8 @@ iai-callgrind.workspace = true
 
 [features]
 rayon = ["dep:rayon"]
-
 default = ["rayon"]
+no_std = []
```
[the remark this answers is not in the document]
MauroToscano: Done

ensure-no_std/src/main.rs
```diff
@@ -0,0 +1,21 @@
...
+#![no_main]
+
+//use core::panic::PanicInfo;
+
+// /// This function is called on panic.
+//#[panic_handler]
+//fn panic(_info: &PanicInfo) -> ! {
+//    loop {}
+//}
+
```
pablodeymo, line 11:
```suggestion
```

pablodeymo (approved):
great work! 👌

gabrielbosio (approved):

.github/workflows/ci.yaml
```diff
@@ -21,11 +21,18 @@ jobs:
...
-          command: check
+        run: cargo check
+
+      - name: Run cargo check for math with no-std
+        run: cargo check --package lambdaworks-math --no-default-features
+
+      - name: Run cargo build ensure-no_std crate
+        run: |
+          cd ensure-no_std
+          cargo build
```
MegaRedHand, line 35: Does the workflow keep the current dir? In that case this should include a `cd ..` or move this step to the end of it.

math/src/fft/errors.rs
```diff
@@ -9,8 +9,8 @@ use lambdaworks_gpu::cuda::abstractions::errors::CudaError;
 
 #[derive(Debug, Error)]
 pub enum FFTError {
-    #[error("Could not calculate {1} root of unity")]
-    RootOfUnityError(String, u64),
+    #[error("Could not calculate root of unity")]
```
MegaRedHand, line 12: Would this be?
```suggestion
    #[error("Could not calculate {0} root of unity")]
```

math/src/lib.rs
```diff
@@ -1,10 +1,17 @@
+#![cfg_attr(feature = "no_std", no_std)]
```
MegaRedHand, line 1:
```suggestion
#![cfg_attr(not(feature = "std"), no_std)]
```

MegaRedHand (approved):
LGTM. Left some comments.

.github/workflows/ci.yaml
MauroToscano: Yes it does

math/src/fft/errors.rs
MauroToscano: No, this is not a mistake, I'm limiting information because I removed thiserror, and it doesn't matter a lot

MauroToscano: Yes, we took a look at that library but it wasn't widely used. So we avoided it for security reasons

MauroToscano:
To Bytes is breaking the CI, I'm fixing it

Issue #445, opened 2023-06-22
Crash when computing `(cairo_prime - 1).sqrt()`


Juan-M-V:
Steps to reproduce:
Write a program that attempts to calculate the square root of `-cairo_prime`
```rust
use lambdaworks_math::field::{
    fields::fft_friendly::stark_252_prime_field::Stark252PrimeField, 
    element::FieldElement
};

fn main() {
    let prime_prev = -FieldElement::<Stark252PrimeField>::one();
    let _ = prime_prev.sqrt();
}
```
Crashes with: thread 'main' panicked at 'attempt to shift left with overflow'

PR #446, opened 2023-06-22
Fix sqrt overflow for moduli with large 2-adicity

schouhy:
# Fix sqrt overflow for moduli with large 2-adicity

## Description
This PR solves Issue #445.

The Tonelli-Shanks algorithm attempts to raise a field element to the power `2^(m-i-1)`. The value` m-i-1` can be of the order of the 2-adicity of the field. At the moment this is computed as `c.pow(1_usize << (m - i - 1))`, which overflows for fields with 2-adicity larger than 128. This is the case for `Stark252PrimeField`.

This PR solves this by avoiding the call to `pow` and replaces it with successive calls to `square`.

## Type of change

- [x] Bug fix

### DIFF
--- a/math/src/field/traits.rs
+++ b/math/src/field/traits.rs
@@ -213,7 +213,10 @@ pub trait IsPrimeField: IsField {
                 i + 1
             };
 
-            let b = Self::pow(&c, 1usize << (m - i - 1));
+            let mut b = c.clone();
+            for _ in 0..(m - i - 1) {
+                b = Self::square(&b);
+            }
 
             c = Self::mul(&b, &b);
             x = Self::mul(&x, &b);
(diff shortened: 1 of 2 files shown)


math/src/field/traits.rs
```diff
@@ -213,7 +213,10 @@ pub trait IsPrimeField: IsField {
                 i + 1
             };
 
-            let b = Self::pow(&c, 1usize << (m - i - 1));
+            let mut b = c.clone();
+            for _ in 0..(m - i - 1) {
+                b = Self::square(&b);
+            }
```
entropidelic, line 219: small nit :)
```rust
 let b = (0..(m - i - 1)).fold(c, |acc, _| Self::square(&acc));
```

entropidelic (approved):

pablodeymo (approved):
LGTM

Oppen (approved):

PR #448, opened 2023-06-23
Implement Serialization for FieldElement

matias-gonz:


MauroToscano:
`from_representative`is not needed, this a `new`, and `FieldElement` serialization isn't needed either. We just need to serialize proofs.

PR #451, opened 2023-06-26
Sha3 512 tree and merkle tree backends refactor

MauroToscano:
# Sha3 512 and MerkleTree Backend

## Description

Add sha3 512, and moved all the backends to one folder

## Type of change

- [x] New feature

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,0 @@
-{
-    "rust-analyzer.linkedProjects": [
-        "./crypto/Cargo.toml"
-    ]
-}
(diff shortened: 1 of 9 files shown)


schouhy (approved):
LGTM. I think that it could be confusing for newcomers that there are things called  `Sha3_XXXTree` and `MerkleTree` and that they are intended to work together. Maybe I would name the the backends `Sha3_XXXBackend` instead, just to emphasize their role.

pablodeymo (approved):
LGTM

diegokingston (approved):

PR #459, opened 2023-06-26
Fix cuda feature

gabrielbosio:
Currently, the clippy step is failing. This PRs attempts to fix the cuda job.


math/src/fft/gpu/cuda/state.rs
Oppen, line 1: Could you redo this by using `git mv gpu/src/cuda/abstractions/state.rs math/src/fft/gpu/cuda/state.rs` instead? It'll make it easier to review if there are actual changes in the file.

Oppen (approved):

MegaRedHand (approved):
👍

entropidelic (approved):

PR #478, opened 2023-07-04
FFT with precomputed twiddle factors

gabrielbosio:


Oppen (approved):

MauroToscano:
This looks great, but it's stale, we can go back at it when needed

PR #484, opened 2023-07-05
From Hex for PrimeField Elements

MauroToscano:
# From Hex for PrimeField

## Description

Add from_hex for PrimeField elements, so it's avaible for generic functions that uses PrimeFields

## Type of change

- [x] New feature

### DIFF
--- a/math/src/field/element.rs
+++ b/math/src/field/element.rs
@@ -416,7 +416,7 @@ impl<F: IsPrimeField> FieldElement<F> {
     /// Returns an `CreationError::InvalidHexString`if the value is not a hexstring
     pub fn from_hex(hex_string: &str) -> Result<Self, CreationError> {
         Ok(Self {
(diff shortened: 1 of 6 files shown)


diegokingston (approved):

math/src/field/element.rs
```diff
@@ -411,6 +411,14 @@ impl<F: IsPrimeField> FieldElement<F> {
     pub fn legendre_symbol(&self) -> LegendreSymbol {
         F::legendre_symbol(&self.value)
     }
+
+    /// Creates a `FieldElement` from a hexstring. It can contain `0x` or not.
+    /// Returns an `CreationError::InvalidHexString`if the value is not a hexstring
```
schouhy, line 416: I think it also returns the CreationError when the hex is too large to fit in a u64.

schouhy (approved):

Oppen (approved):

PR #491, opened 2023-07-06
Add alloc feture (nostd + alloc) for poly and ByteConversion

MauroToscano:
# Add alloc feature for poly (nostd + alloc)

## Description

- Add `alloc` target in Math for wasm, a less strict no-std that allows allocs
- Made `poly` compatible with `no-std` with `alloc`
- Made `batch_inv` compatible with `no-std` with `alloc`
- Enabled `ByteConversion` with `no-std` with `alloc`

## Type of change

- [x] New feature


pablodeymo (approved):
🥇

PR #504, opened 2023-07-11
Fix CUDA and Metal FFT reaching block/threadgroup limit

### LINKED ISSUE #449: FFT evaluation on metal fails on a prover test
There's a test from the Cairo prover that's making the FFT evaluator return a different array when metal is enabled
 In [this CI run](https://github.com/lambdaclass/lambdaworks_cairo_prover/actions/runs/5360751135/jobs/9725998402?pr=102), the job for metal tests failed but the job that runs tests with metal disabled passed.

xqft:
# Fix CUDA and Metal FFT reaching block/threadgroup limit

## Description

CUDA and Metal FFT fail when executing with an input larger than `2^10` elements. This is because both Metal and CUDA have a threads per block/threadgroup limit of 1024 threads.

Closes #449 

## Type of change

- Bug fix

## Checklist
- [x] Linked to Github Issue
- [x] Unit tests added


math/src/fft/gpu/cuda/ops.rs
```diff
@@ -27,10 +27,11 @@ where
 
     let order = input.len().trailing_zeros();
     for stage in 0..order {
-        let group_count = 1 << stage;
-        let group_size = input.len() / group_count;
+        const WARP_SIZE: usize = 32;
+        let block_size = WARP_SIZE;
+        let block_count = (input.len() + block_size) / block_size;
```
MegaRedHand, line 32: This should make it so we don't have an additional empty block/threadgroup when `input` is divisible by `block_size`:
```suggestion
        let block_count = (input.len() + block_size - 1) / block_size;
```

math/src/fft/gpu/cuda/ops.rs
```diff
@@ -116,6 +117,22 @@ mod tests {
         }
     }
 
+    // May want to modify the order constant, takes ~5s to run on a M1.
```
MegaRedHand, line 120: I think this doesn't apply here

math/src/fft/gpu/metal/ops.rs
```diff
@@ -34,17 +34,18 @@ pub fn fft<F: IsFFTField>(
...
         let order = input.len().trailing_zeros();
         for stage in 0..order {
-            let group_count = 1 << stage;
-            let group_size = input.len() as u64 / group_count;
+            command_encoder.set_bytes(2, mem::size_of_val(&stage) as u64, void_ptr(&stage));
 
-            let threadgroup_size = MTLSize::new(group_size / 2, 1, 1);
-            let threadgroup_count = MTLSize::new(group_count, 1, 1);
-            command_encoder.dispatch_thread_groups(threadgroup_count, threadgroup_size);
+            let grid_size = MTLSize::new(input.len() as u64 / 2, 1, 1); // one thread per butterfly
```
MegaRedHand, line 44: Is it ok to divide by 2 always? Shouldn't it be `pipeline.thread_execution_width()`?

math/src/fft/gpu/cuda/ops.rs
xqft: oops, copy-paste err, fixed!

math/src/fft/gpu/metal/ops.rs
xqft: yep, input length will always be a power of two. See [here](https://github.com/lambdaclass/lambdaworks/blob/8fcd64f1db99a249a8bdfa84bff9731730455ca9/math/src/fft/gpu/metal/ops.rs#L24-L26)

xqft: `grid_size` is the total amount of threads that will be spawned, not the threadgroup size

math/src/fft/gpu/cuda/ops.rs
xqft: you are right! fixed

math/src/fft/gpu/metal/ops.rs
MegaRedHand: Yeah, I got a bit confused with the current implementation. Really weird we didn't catch this before 😅

math/src/fft/gpu/cuda/state.rs
```diff
@@ -235,16 +226,14 @@ impl<F: IsField> CalcTwiddlesFunction<F> {
         }
     }
 
-    pub(crate) fn launch(&mut self, group_size: usize) -> Result<(), CudaError> {
-        let grid_dim = (1, 1, 1); // in blocks
-        let block_dim = (group_size as u32, 1, 1);
+    pub(crate) fn launch(&mut self, count: usize) -> Result<(), CudaError> {
+        const WARP_SIZE: usize = 32;
```
MegaRedHand, line 230: We should declare this in only one place

MegaRedHand (approved):

diegokingston (approved):

gabrielbosio (approved):

Issue #486, opened 2023-07-05
Fix `from_hex_unchecked` implementation for field elements with montgomery backend


entropidelic:
```rust
impl<M, const NUM_LIMBS: usize> FieldElement<MontgomeryBackendPrimeField<M, NUM_LIMBS>>
where
    M: IsModulus<UnsignedInteger<NUM_LIMBS>> + Clone + Debug,
{
    /// Creates a `FieldElement` from a hexstring. It can contain `0x` or not.
    /// # Panics
    /// Panics if value is not a hexstring
    pub const fn from_hex_unchecked(hex: &str) -> Self {
        let integer = UnsignedInteger::<NUM_LIMBS>::from_hex_unchecked(hex);
        Self {
            value: MontgomeryAlgorithms::cios(
                &integer,
                &MontgomeryBackendPrimeField::<M, NUM_LIMBS>::R2,
                &M::MODULUS,
                &MontgomeryBackendPrimeField::<M, NUM_LIMBS>::MU,
            ),
        }
    }
}
```
here the documentation of the function states that the number can contain the `0x` prefix or not, but then `UnsignedInteger::<NUM_LIMBS>::from_hex_unchecked(hex);` is called, and that function has the following documentation

```rust
 /// Creates an `UnsignedInteger` from a hexstring
    /// # Panics
    /// Panics if value is not a hexstring. Shouldn't start with `0x`
    pub const fn from_hex_unchecked(value: &str) -> Self {
```

so there is an inconsistency. Either the documentation must be changed or the functionality to strip the prefix must be added

startup-dreamer:
I think for the sake of consistancy we should modifie `from_hex_unchecked` for `UnsignedInteger` as `from_hex` also states 

https://github.com/lambdaclass/lambdaworks/blob/8fcd64f1db99a249a8bdfa84bff9731730455ca9/math/src/unsigned_integer/element.rs#L420-L421
for example this check https://github.com/lambdaclass/lambdaworks/blob/8fcd64f1db99a249a8bdfa84bff9731730455ca9/math/src/unsigned_integer/element.rs#L425-L432
should be used in `from_hex_unchecked` instead of `from_hex` it will make functions more consistance and we can than update the docs of `UnsignedInteger's` `from_hex_unchecked.`

Thought's @Oppen @MauroToscano

PR #510, opened 2023-07-16
Fix: from_hex_unchecked implementation for field elements

startup-dreamer:
# TITLE

## Description

fixed the `from_hex_unchecked` for `FieldElement` which can now create field elements with '0x' hex strings. added the required tests.

## Type of change
Fix: #486 

- [x] Bug fix
- [x] Optimization

## Checklist
- [x] Linked to Github Issue
- [x] Unit tests added


Oppen (approved):

MauroToscano (approved):

math/src/unsigned_integer/element.rs
```diff
@@ -440,23 +440,29 @@ impl<const NUM_LIMBS: usize> UnsignedInteger<NUM_LIMBS> {
...
-                    panic!("Malformed hex expression.")
-                }
+
+        let value_bytes = value.as_bytes();
+
+        // Remove "0x" if it's at the beginning of the string
+        let mut i = 0;
+        if value_bytes.len() > 2 && value_bytes[0] == b'0' && value_bytes[1] == b'x' {
+            i = 2;
+        }
```
pablodeymo, line 456: I think it's better to use `starts_with` in this case:
https://doc.rust-lang.org/std/primitive.slice.html#method.starts_with

startup-dreamer: @pablodeymo As `from_hex_unchecked` is a const function it will not call a non-const function. so is there any other way we can implement this

Oppen: Maybe:
```rust
if value_bytes.len() > 2 && value_bytes[..2] == b'0x' {
```

startup-dreamer: I don't think this is going to work either compiler is giving this result.
```rust
    |   
454 |         if value_bytes.len() > 2 && value_bytes[..2] == b"0x" {
    |                                                      ^^ no implementation for `[u8] == &[u8; 2]`
```
alternatively if i try something like this 
```rust
454 |         if value_bytes.len() > 2 && value_bytes[..2] == [b'0', b'x'] 
```
it won't work as stated previously that `from_hex_unchecked` can't call a non const operator.

startup-dreamer: no it still isn't working.

math/src/field/element.rs
```diff
@@ -661,6 +661,13 @@ mod tests {
         assert_eq!(FE::from_hex("1a").unwrap(), FE::from(26))
     }
 
+    #[test]
+    fn from_hex_unchecked_zero_x_1a_is_26_for_stark252_prime_field_element() {
+        type F = Stark252PrimeField;
+        type FE = FieldElement<F>;
+        assert_eq!(FE::from_hex_unchecked("0x1a"), FE::from(26)
```
MauroToscano, line 668:
```suggestion
        assert_eq!(FE::from_hex_unchecked("0x1a"), FE::from(26))
```

PR #513, opened 2023-07-18
Impl Bandersnatch curve

0xKitetsu-smdk:
# Bandersnatch Curve

## Description
- Bandersnatch, a curve over BLS12-381 scalar field, with glv has efficient scalar multiplication over jubjub curve
- This PR implements Twisted Edward form of Bandersnatch curve
- This PR doesn't implement GLV, only the curve.
- You can find the reference here: https://eprint.iacr.org/2021/1152.pdf

## Type of change

- [x] New feature

## Checklist
- [ ] Linked to Github Issue
- [x] Unit tests added
- [ ] Benchmarks added/run


dragan2234:
Copy-pasting the comment in case we or somebody continues working on this:

Some conclustion:

As a next step I thought creating a test for adding a point of infinity to some point, can be a generator. That should result in the same point. So like 56 + 0 = 56 if we think in integers.

But there is a problem with lambdaworks affine and/or projective traits. Neither one of those have an example of the representations of the point of infinity. I tried creating one but then got the assert error on this line:
https://github.com/lambdaclass/lambdaworks/blob/main/math/src/elliptic_curve/point.rs#L44


Another task that we have is having a GLV endomorphism implemented which is a big task, but when trying to decompose it into smaller tasks, we realized that as first step we can implement scalar decomposition which uses "Babai's nearest plane algorithm" implemented here:
https://github.com/zhenfeizhang/bandersnatch/blob/main/bandersnatch/src/curves/glv.rs#L97

This looks trivial, but uses bigint and lambdaworks doesn't have bigint (where arkworks has it's own bigint implementation, although i know there exist rust bigint libraries, arkworks uses their own).

math/src/elliptic_curve/edwards/curves/bandersnatch/curve.rs
```diff
@@ -0,0 +1,129 @@
...
+        Self::PointRepresentation::new([
+            FieldElement::<Self::BaseField>::new_base(
+                "29C132CC2C0B34C5743711777BBE42F32B79C022AD998465E1E71866A252AE18",
+            ),
+            FieldElement::<Self::BaseField>::new_base(
+                "2A6C669EDA123E0F157D8B50BADCD586358CAD81EEE464605E3167B6CC974166",
+            ),
+            FieldElement::one(),
+        ])
+    }
```
MauroToscano, line 27: The generators don't seem to match. This is in projective coordinates, and the one linked are in affine

MauroToscano: Alright, the conversion from affine to projective is good

diegokingston (approved):

math/src/elliptic_curve/edwards/curves/bandersnatch/field_extension.rs
MauroToscano, line 1: Should be field, since it's not an extension

schouhy (approved):

MauroToscano (approved):

dragan2234:
@MauroToscano Thanks for merging! We could add a "collaborator-friendly" follow-up issue on the repo regarding the GLV endomorphism (I've put some notes on the comment above) because AFAIU that's the main purpose of this curve and what makes it different from jubjub ( https://github.com/zkcrypto/jubjub ) https://github.com/zhenfeizhang/bandersnatch#examples

PR #518, opened 2023-08-01
Improve benches

MauroToscano:
# General improvements for the benches

## Description

- Fix seed so results are always the same
- Fixed all benchmarks but sqrt to use 10000 operations with random elements
- Add the amount of operations made to the text of the benchmarks, so it's clear the time measured it's not one operation
- Updated the text that shows which arkworks version is used 

## Type of change

- [x] New feature


diegokingston (approved):

pablodeymo (approved):
LGTM

entropidelic (approved):

schouhy (approved):

PR #520, opened 2023-08-02
Optimize add to avoid overflow check

schouhy:
# Optimize add to avoid overflow check

## Type of change
- [x] Optimization

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run

### DIFF
--- a/math/src/field/fields/montgomery_backed_prime_fields.rs
+++ b/math/src/field/fields/montgomery_backed_prime_fields.rs
@@ -41,6 +41,7 @@ where
         &M::MODULUS,
         &Self::MU,
     );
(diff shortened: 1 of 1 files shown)


Oppen:
Two suggestions:
1. Make `modulus_has_one_spare_bit` `#[inline(always)]`;
2. Evaluate to a `const` before the `if` guard, so we can be sure it's always evaluated at compile time.
Point 2 is because a `const fn` is not intrinsically always evaluated as a constant, but it signals that it _can_ be used to create constants. By storing the result into a `const` you guarantee that, so then the correct branch is known at compile-time to always be taken and thus the check and the other branch get removed by the compiler.

Oppen (approved):

diegokingston (approved):

PR #521, opened 2023-08-07
perf: use overflowing_add for UnsignedInteger::add

Oppen:
This allows using only `u64` variables and gives a bit more information
to the compiler about our intent.

Results on M1 with 16GB:
```
add 1M elements | lambdaworks
                        time:   [7.0109 ms 7.0278 ms 7.0465 ms]
                        change: [-29.850% -29.645% -29.430%] (p = 0.00 < 0.01)
                        Performance has improved.
```

## Checklist
  - [x] Benchmarks added/run


diegokingston (approved):

schouhy (approved):
awesome!

MauroToscano (approved):

PR #526, opened 2023-08-15
feat(proptest): Implement `proptest::Arbitrary` for `UnsignedInteger`

### LINKED ISSUE #414: Add `proptest::Arbitrary` support
In [_cairo-vm_](https://github.com/lambdaclass/cairo-vm/blob/c2b7f1818db110869012f75ad72b9f97f40e9c5d/felt/src/arbitrary.rs#L8) we implement [`proptest::Arbitrary`](https://docs.rs/proptest/latest/proptest/) on our `Felt252` type to use it on proptests. We could move that code to `UnsignedInteger`, to allow lib users to more easily proptest their projects. This also should help reduce the amount of code repetition in _lambdaworks_' tests.

PatStiles:
Closes #414 

Adds an implementation of `proptest::Arbitrary` for `UnsignedInteger`. Refactors proptests to use `any::<UnsignedInteger>()`

### DIFF
--- a/math/Cargo.toml
+++ b/math/Cargo.toml
@@ -11,7 +11,7 @@ license.workspace = true
 thiserror = { version = "1.0", optional = true }
 serde = { version = "1.0", features = ["derive"], optional = true }
 serde_json = { version = "1.0", optional = true }
-proptest = "1.1.0"
+proptest = { version = "1.1.0", optional = true }
 
 # rayon
 rayon = { version = "1.7", optional = true }
@@ -30,12 +30,14 @@ rand = { version = "0.8.5", default-features = false }
 criterion = "0.4"
 const-random = "0.1.15"
 iai-callgrind.workspace = true
+proptest = "1.1.0"
 
 [features]
 rayon = ["dep:rayon"]
 default = ["rayon", "std"]
 std = ["dep:thiserror"]
 lambdaworks-serde = ["dep:serde", "dep:serde_json", "std"]
+proptest = ["dep:proptest"]
 
 # gpu
 metal = [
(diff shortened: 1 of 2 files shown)


PatStiles:
Open to any feedback! Always looking to improve.

math/Cargo.toml
```diff
@@ -11,6 +11,7 @@ license.workspace = true
 thiserror = { version = "1.0", optional = true }
 serde = { version = "1.0", features = ["derive"], optional = true }
 serde_json = { version = "1.0", optional = true }
+proptest = "1.1.0"
```
Oppen, line 14: Please, make this dependency optional and use a feature flag. We don't want to force the dependency for users who aren't interested in testing with arbitrary numbers.

PatStiles: Resolved!

math/src/unsigned_integer/element.rs
```diff
@@ -963,7 +969,8 @@ mod tests_u384 {
 
     proptest! {
         #[test]
-        fn bitand(a in any::<[u64; N_LIMBS]>(), b in any::<[u64; N_LIMBS]>()) {
+        #[cfg(feature = "proptest")]
```
Oppen, line 972: Isn't `proptest` always enabled for tests? In any case, you'd probably save time and code by flagging the whole `proptest` block instead.

PatStiles: lol yeah that makes a ton of sense thank you!

PatStiles: Followed your suggestion for flagging!

MauroToscano (approved):

schouhy (approved):

diegokingston (approved):

Issue #545, opened 2023-09-04
Error instead of panicking on `inv` function for `FieldElement`


MauroToscano:
Inv should not panic. It should be changed to an error

- [ ] Add `InvZeroError`, unwrap where before it would panic
- [ ] Remove unwraps

First task should be done and merged before the second one.

dbejarano820:
I can take this one!

diegokingston:
Solved with https://github.com/lambdaclass/lambdaworks/pull/546

PR #546, opened 2023-09-04
feat: Error instead of panicking on inv function for FieldElement

dbejarano820:
# Error instead of panicking on inv function for FieldElement

## Description
This PR is the first step in closing #545 .
It adds InvZeroError and unwraps where before it would panic

## Type of change
- [X] Optimization

## Checklist
- [X] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run

### DIFF
--- a/math/src/elliptic_curve/short_weierstrass/curves/bls12_381/field_extension.rs
+++ b/math/src/elliptic_curve/short_weierstrass/curves/bls12_381/field_extension.rs
@@ -8,6 +8,7 @@ use crate::{
         },
         fields::montgomery_backed_prime_fields::{IsModulus, MontgomeryBackendPrimeField},
         traits::IsField,
+        errors::FieldError,
     },
     traits::ByteConversion,
 };
@@ -65,14 +66,14 @@ impl IsField for Degree2ExtensionField {
 
     /// Returns the multiplicative inverse of `a`
     /// This uses the equality `(a0 + a1 * t) * (a0 - a1 * t) = a0.pow(2) - a1.pow(2) * Q::residue()`
-    fn inv(a: &Self::BaseType) -> Self::BaseType {
-        let inv_norm = (a[0].pow(2_u64) + a[1].pow(2_u64)).inv();
-        [&a[0] * &inv_norm, -&a[1] * inv_norm]
+    fn inv(a: &Self::BaseType) -> Result<Self::BaseType, FieldError> {
+        let inv_norm = (a[0].pow(2_u64) + a[1].pow(2_u64)).inv().unwrap();
+        Ok([&a[0] * &inv_norm, -&a[1] * inv_norm])
     }
 
     /// Returns the division of `a` and `b`
     fn div(a: &Self::BaseType, b: &Self::BaseType) -> Self::BaseType {
-        Self::mul(a, &Self::inv(b))
+        Self::mul(a, &Self::inv(b).unwrap())
     }
 
     /// Returns a boolean indicating whether `a` and `b` are equal component wise.
--- a/math/src/elliptic_curve/short_weierstrass/curves/bls12_381/field_extension.rs
+++ b/math/src/elliptic_curve/short_weierstrass/curves/bls12_381/field_extension.rs
@@ -2,13 +2,13 @@ use crate::unsigned_integer::element::U384;
 use crate::{
     field::{
         element::FieldElement,
+        errors::FieldError,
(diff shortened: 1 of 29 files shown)


math/src/elliptic_curve/short_weierstrass/curves/bls12_381/field_extension.rs
```diff
@@ -65,14 +66,14 @@ impl IsField for Degree2ExtensionField {
 
     /// Returns the multiplicative inverse of `a`
     /// This uses the equality `(a0 + a1 * t) * (a0 - a1 * t) = a0.pow(2) - a1.pow(2) * Q::residue()`
-    fn inv(a: &Self::BaseType) -> Self::BaseType {
-        let inv_norm = (a[0].pow(2_u64) + a[1].pow(2_u64)).inv();
-        [&a[0] * &inv_norm, -&a[1] * inv_norm]
+    fn inv(a: &Self::BaseType) -> Result<Self::BaseType, FieldError> {
+        let inv_norm = (a[0].pow(2_u64) + a[1].pow(2_u64)).inv().unwrap();
```
Oppen, line 70: Is `a[0].pow(2_u64) + a[1].pow(2_u64)` guaranteed to never be `0` in the field? If so, please add a `SAFETY:` comment explaining why. Otherwise, this should bubble up the error.

math/src/field/element.rs
```diff
@@ -39,7 +40,7 @@ impl<F: IsField> FieldElement<F> {
         for i in 1..count {
             prod_prefix.push(&prod_prefix[i - 1] * &numbers[i]);
         }
-        let mut bi_inv = prod_prefix[count - 1].inv();
+        let mut bi_inv = prod_prefix[count - 1].inv().unwrap();
```
Oppen, line 43: Same issue. We could settle for postponing in this case if needed though, to keep the splash damage more contained.

MauroToscano:
Good PR, we should propagate the error through the invert functions. The ones calling them can unwrap for the time being if needed, as to limit the scope of this issue. 

As a note, you should run `make clippy` and apply the changes. You can usually apply them with cargo `clippy --fix`. You should also run `cargo fmt --all`. The only task of the CI that can fail until we fix it is the NVIDIA one, since we are having an issue with the server.

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/field_extension.rs
dbejarano820: Sounds good

math/src/field/element.rs
```diff
@@ -694,7 +695,7 @@ mod tests {
         fn test_inplace_batch_inverse_returns_inverses(vec in field_vec(10)) {
             let input: Vec<_> = vec.into_iter().filter(|x| x != &FieldElement::<Stark252PrimeField>::zero()).collect();
             let mut inverses = input.clone();
-            FieldElement::inplace_batch_inverse(&mut inverses);
+            let _ = FieldElement::inplace_batch_inverse(&mut inverses);
```
Oppen, line 698: Let's `unwrap` here. A `panic` is better than continuing with a inverses that aren't inverses.

math/src/polynomial.rs
```diff
@@ -66,7 +66,7 @@ impl<F: IsField> Polynomial<FieldElement<F>> {
             }
         }
 
-        FieldElement::inplace_batch_inverse(&mut denominators);
+        let _ = FieldElement::inplace_batch_inverse(&mut denominators);
```
Oppen, line 69: `unwrap` here as well.

provers/plonk/src/prover.rs
```diff
@@ -497,7 +497,7 @@ where
             .collect();
 
         let mut zh_eval = zh.evaluate_offset_fft(1, Some(degree), offset).unwrap();
-        FieldElement::inplace_batch_inverse(&mut zh_eval);
+        let _ = FieldElement::inplace_batch_inverse(&mut zh_eval);
```
Oppen, line 500: And here.

math/src/field/element.rs
dbejarano820: Ok will do

diegokingston (approved):

MauroToscano (approved):

Oppen (approved):

schouhy (approved):

Issue #539, opened 2023-09-04
Add Babybear field to pre defined fields


MauroToscano:
This task consists of three tasks:

- [x] Add Babybear field to pre defined fields on : math/src/field/fields/fft_friendly, using the right modulus.
- [x] Implement IsPrimeField for BabyBear
- [x] Make it FFT friendly
- [ ] Add the degree six and eight extension

For reference, this is the 31 bit used by Risc Zero, as described [here](https://dev.risczero.com/proof-system-in-detail.pdf):

This field is suitable to be used with the MontgomeryBackend. It can be done with a small MontgomeryBackend of 32 bits later. But if it's not available when doing this issue, it can be done using the one with u64 limbs.

Constant for the modulus is:

2^31 - 2^27 + 1

For first time contributors, we recommend doing task 1, then 2, then 3, and skip task 4. 

It may be useful to finish [this issue](https://github.com/lambdaclass/lambdaworks/issues/538) before three, but it's not mandatory.

mdvillagra:
Do you want the babybear field to be in a similar vein as https://github.com/lambdaclass/lambdaworks/blob/main/math/src/field/fields/p448_goldilocks_prime_field.rs?

mdvillagra:
> Since there's code for the FFT/NTT, there should be a test to check if it's working properly. It can be submitted as a package of IsFFt with tests in another PR, or fixed here. The IsFieldpart seems right.

@MauroToscano which tests? which code are you referring to?

mdvillagra:
@MauroToscano is it really necessary to implement the `isPrime` trait? It is already implemented for `MontgomeryBackendPrimeField` [here](https://github.com/lambdaclass/lambdaworks/blob/8e0438adb11cdbcb404116a63f51970369102c62/math/src/field/fields/montgomery_backed_prime_fields.rs#L277) and babybear already uses `MontgomeryBackendPrimeField`.

PR #549, opened 2023-09-05
Add Babybear field

mdvillagra:
# Babybear field implementation

## Description

Add Babybear field with arithmetic operations as requested on issue #539. The field uses `MontgomeryBackendPrimeField` and implements the same traits as in `Stark252PrimeField`. The field elements are encoded using type `u64` considering that Babybear only requires 31 bits.

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Optimization

## Checklist
- [x] Linked to Github Issue
- [x] Unit tests added
  - [ ] Benchmarks added/run


MauroToscano (changes requested):
Babybear Field should be using the `MontgomeryBackendPrimeField`. The code should set the right constants using that backend. See `Stark252PrimeField`as an example

mdvillagra:
> Babybear Field should be using the `MontgomeryBackendPrimeField`. The code should set the right constants using that backend. See `Stark252PrimeField`as an example

What about the field operations? Do I still leave them there? Or just implement the modulus? The file `Stark252PrimeField` doesn't have any field operations implemented.

math/src/field/fields/fft_friendly/babybear.rs
```diff
@@ -0,0 +1,102 @@
+use crate::{
+    field::{
+        element::FieldElement,
+        fields::montgomery_backed_prime_fields::{IsModulus, MontgomeryBackendPrimeField},
+        //traits::IsFFTField,
+    },
+    unsigned_integer::element::UnsignedInteger,
+};
+
+pub type U64 = UnsignedInteger<1>;
```
schouhy, line 10: Please move this to `unsigned_integer/element.rs` ([here](https://github.com/lambdaclass/lambdaworks/blob/29af640a8e4e22301e7866981f156281d03ab51a/math/src/unsigned_integer/element.rs#L22-L24)). So all unsigned integer type aliases are together.

math/src/field/fields/fft_friendly/babybear.rs
```diff
@@ -0,0 +1,102 @@
...
+pub type U64PrimeField<T> = MontgomeryBackendPrimeField<T, 1>;
```
schouhy, line 11: There's already a `U64PrimeField` ([here](https://github.com/lambdaclass/lambdaworks/blob/29af640a8e4e22301e7866981f156281d03ab51a/math/src/field/fields/u64_prime_field.rs#L15)). 
@MauroToscano @Oppen I guess there's no need to have both u64primefields if the one with the Montgomery backend is more performant. Maybe we should create an issue to benchmark them and remove the slow one.
For the time being I would change the name of this type alias to `U64MontgomeryBackendPrimeField` or something along those lines. To avoid confusion.

schouhy (changes requested):
Nice! Thanks for the PR!
I left a minor request change to move and rename stuff.

math/src/field/fields/fft_friendly/babybear.rs
MauroToscano: We should review this. But let's do it ourselves, and discuss it outside this PR.

math/src/field/fields/fft_friendly/babybear.rs
MauroToscano: This a nice idea

math/src/field/fields/fft_friendly/babybear.rs
MauroToscano: The change of name would be nice though

mdvillagra:
guys, I don't know why another PR was created about this issue.

schouhy:
no worries

mdvillagra:
I did some clean up on my fork, and now it seems to be fine. I didn't noticed but I was working on another branch. Now I pushed into my main branch.

schouhy:
Cool. Looks like the lint check is failing. That's prolly solved by running `cargo fmt`.

schouhy (approved):
Awesome. Thanks!

diegokingston (approved):

MauroToscano (changes requested):
Since there's code for the FFT/NTT, there should be a test to check if it's working properly. It can be submitted as a package of `IsFFt` with tests in another PR, or fixed here. The `IsField`part seems right. 

Either add the tests, or remove the FFT part

mdvillagra:
> Since there's code for the FFT/NTT, there should be a test to check if it's working properly. It can be submitted as a package of `IsFFt` with tests in another PR, or fixed here. The `IsField`part seems right.
> 
> Either add the tests, or remove the FFT part

I will delete the FFT part right now and add it in another PR.

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "githubPullRequests.ignoredPullRequestBranches": [
+        "main"
+    ]
+}
```
dbejarano820, line 5: Do we really want to merge this?

mdvillagra: probably is better to just delete it right?

mdvillagra: I deleted the file

math/src/field/fields/fft_friendly/babybear.rs
```diff
@@ -0,0 +1,101 @@
...
+    fn partial_cmp(&self, other: &Self) -> Option<core::cmp::Ordering> {
+        self.representative().partial_cmp(&other.representative())
+    }
+}
+
+impl Ord for FieldElement<Babybear31PrimeField> {
+    fn cmp(&self, other: &Self) -> core::cmp::Ordering {
+        self.representative().cmp(&other.representative())
+    }
+}
```
MauroToscano, line 43: This isn't needed, but it's ok, we can document this later and remove if needed. Doesn't hurt either

MauroToscano (approved):

dbejarano820 (approved):

PR #555, opened 2023-09-12
todo: implement getters in plonk setup

dbejarano820:
# Getters in plonk setup

## Description
Browsing through the code I noticed a simple ToDo task which is what this PR solves.
I'll complete the other ToDo in a separate PR which involves removing unwraps in the `from_constraint_system` function in this same file.

## Type of change
- [X] ToDo task

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run

### DIFF
--- a/provers/plonk/src/setup.rs
+++ b/provers/plonk/src/setup.rs
@@ -29,7 +29,7 @@ impl<F: IsField> Witness<F> {
             c: abc[2 * n..].to_vec(),
         }
     }
-    
+
     // Getters
     pub fn get_a(&self) -> &Vec<FieldElement<F>> {
         &self.a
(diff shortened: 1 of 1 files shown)


diegokingston (approved):

MauroToscano:
We don't use getters in the library. And the attributes are public. Seems like an odd TO DO. I'd avoid adding them. Let me check with the rest of the team.

MauroToscano:
For the time being we shouldn't do this. If you want to remove the comment with the TO DO we can do that. 

We will include getters for private attributes, but as you have seen most of the repo functions and structs are public. 

Before the main release we will review the visibility of functions making pub (crate) most of them, and then expose getters as needed. This will avoid breaking changes for people using the library.

PR #557, opened 2023-09-18
Add Ed448-Goldilocks elliptic curve

### LINKED ISSUE #37: Add support for Goldilocks elliptic curve

techiepriyansh:
Adds support for the Goldilocks elliptic curve (resolves #37).

Also fixes an equality testing bug in the Goldilocks prime field. The bug was caused due to comparing weakly reduced values in the limbs. So, introduce a stronger reduction that converts to the canonical form (each of the 8 `U56x8` limbs have value < `2^56` and total value represented by the limbs < `p = 2^448 - 2^224 - 1` after this reduction) and apply that before comparing the limbs for equality.

### DIFF
--- a/math/src/elliptic_curve/edwards/curves/ed448_goldilocks.rs
+++ b/math/src/elliptic_curve/edwards/curves/ed448_goldilocks.rs
@@ -1,0 +1,99 @@
+use crate::{
+    elliptic_curve::{
+        edwards::{point::EdwardsProjectivePoint, traits::IsEdwards},
+        traits::IsEllipticCurve,
+    },
+    field::{element::FieldElement, fields::p448_goldilocks_prime_field::P448GoldilocksPrimeField},
+};
+
+#[derive(Debug, Clone)]
+pub struct Ed448Goldilocks;
+
+impl IsEllipticCurve for Ed448Goldilocks {
+    type BaseField = P448GoldilocksPrimeField;
+    type PointRepresentation = EdwardsProjectivePoint<Self>;
+
+    /// Taken from https://www.rfc-editor.org/rfc/rfc7748#page-6
+    fn generator() -> Self::PointRepresentation {
+        Self::PointRepresentation::new([
+            FieldElement::<Self::BaseField>::from_hex("4f1970c66bed0ded221d15a622bf36da9e146570470f1767ea6de324a3d3a46412ae1af72ab66511433b80e18b00938e2626a82bc70cc05e").unwrap(),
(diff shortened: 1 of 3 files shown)


math/src/elliptic_curve/edwards/curves/ed448_goldilocks.rs
```diff
@@ -0,0 +1,102 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{
+        cyclic_group::IsGroup, elliptic_curve::traits::EllipticCurveError,
+        field::element::FieldElement,
+    };
+
+    #[allow(clippy::upper_case_acronyms)]
+    type FEE = FieldElement<P448GoldilocksPrimeField>;
```
schouhy, line 45: We usually use `FEE` as short for Field Element Extension. Since this is a prime field, I would rename this to `FE`

math/src/field/fields/p448_goldilocks_prime_field.rs
```diff
@@ -209,12 +247,18 @@ impl U56x8 {
     }
 }
 
+impl FieldElement<P448GoldilocksPrimeField> {
+    pub fn new_base(a_hex: &str) -> Self {
+        Self::new(U56x8::from(a_hex))
+    }
+}
+
```
schouhy, line 255: This should be named `from_hex`. Since this field is prime and `from_hex` is one of the methods of the trait `IsPrimeField`, it seems like the most sensible thing to do here is implement that trait for `P448GoldilocksPrimeField`. Use `U448` as `RepresentativeType`.

schouhy (changes requested):
Awesome! Thanks for the PR. 
I left some comments that I think will improve it. Let me know what you think.

math/src/elliptic_curve/edwards/curves/ed448_goldilocks.rs
techiepriyansh: Done.

math/src/field/fields/p448_goldilocks_prime_field.rs
techiepriyansh: Done.

schouhy (approved):

diegokingston (approved):

pablodeymo (approved):
LGTM

Issue #543, opened 2023-09-04
Add subgroup check to pairings


MauroToscano:
The inputs of the [pairing function](https://github.com/lambdaclass/lambdaworks/blob/6ad2e4620128962657d58a667e1b26d6e948d461/math/src/elliptic_curve/short_weierstrass/curves/bls12_381/pairing.rs#L20-L41) need to be checked to belong to the $r$-torsion of the curve. The naive implementation would be to check that $rG$ is the point at infinity. There are better approaches. See [here](https://hackmd.io/@yelhousni/bls12_subgroup_check).

It should be added to the pairing module of the curve

PatStiles:
I'll take a shot at this

RajeshRk18:
I have just done with adding efficient subgroup checks (using endomorphism) for G1 and G2 here: https://github.com/RajeshRk18/lambdaworks/tree/add/subgroup-checks

MauroToscano:
This was solved with #649

PR #559, opened 2023-09-19
feat(pairing): BLS12-381 Subgroup Check

PatStiles:
# Add Naive Subgroup Check to pairing

## Description

Addresses part of #543 by adding a naive implementation of a pairing subgroup check leveraging the check_point_is_in_subgroup() from compression. To accomodate this check the following was added:

- The trait implementation for IsPairing was 
- A new Error enum was added to [src/math/elliptic_curve/short_weierstrauss/curves/error.rs](https://github.com/lambdaclass/lambdaworks/blob/main/math/src/elliptic_curve/short_weierstrass/errors.rs)
- Made check_point_is_in_subgroup() accept ShortWeierstrassProjectivePoint generic over IsShortWeierstrauss.

@MauroToscano @diegokingston I wasn't sure if using a result was the best practice but it seemed simplest at the time. Let me know if you have any suggestions?

- [x] New feature

## Checklist
- [x] Linked to Github Issue
- [x] Unit tests added

### DIFF
--- a/provers/cairo/Cargo.toml
+++ b/provers/cairo/Cargo.toml
@@ -34,7 +34,7 @@ serde_cbor = { version = "0.11.1" }
 # For cli
 clap = { version = "4.4.6", features = ["derive"], optional = true }
 # Parallelization crates
-rayon = { version = "1.7.0", optional = true }
+rayon = { version = "1.8.0", optional = true }
 
 # wasm
 wasm-bindgen = { version = "0.2", optional = true }
--- a/provers/stark/Cargo.toml
+++ b/provers/stark/Cargo.toml
@@ -23,7 +23,7 @@ num-integer = "0.1.45"
 itertools = "0.11.0"
 
 # Parallelization crates
-rayon = { version = "1.7.0", optional = true }
+rayon = { version = "1.8.0", optional = true }
 
 # wasm
 wasm-bindgen = { version = "0.2", optional = true }
--- a/provers/stark/src/grinding.rs
+++ b/provers/stark/src/grinding.rs
@@ -1,3 +1,5 @@
+#[cfg(feature = "parallel")]
+use rayon::prelude::{IntoParallelIterator, ParallelIterator};
 use sha3::{Digest, Keccak256};
 
 const PREFIX: [u8; 8] = [0x01, 0x23, 0x45, 0x67, 0x89, 0xab, 0xcd, 0xed];
@@ -38,8 +40,16 @@ pub fn is_valid_nonce(seed: &[u8; 32], nonce: u64, grinding_factor: u8) -> bool 
 pub fn generate_nonce(seed: &[u8; 32], grinding_factor: u8) -> Option<u64> {
     let inner_hash = get_inner_hash(seed, grinding_factor);
     let limit = 1 << (64 - grinding_factor);
-    (0..u64::MAX)
-        .find(|&candidate_nonce| is_valid_nonce_for_inner_hash(&inner_hash, candidate_nonce, limit))
+
+    #[cfg(not(feature = "parallel"))]
+    return (0..u64::MAX).find(|&candidate_nonce| {
+        is_valid_nonce_for_inner_hash(&inner_hash, candidate_nonce, limit)
+    });
+
+    #[cfg(feature = "parallel")]
+    return (0..u64::MAX).into_par_iter().find_any(|&candidate_nonce| {
+        is_valid_nonce_for_inner_hash(&inner_hash, candidate_nonce, limit)
+    });
 }
 
 /// Checks if the leftmost 8 bytes of `Hash(inner_hash || candidate_nonce)` are less than `limit`


MauroToscano:
An error is how it's done in Rust, it's good

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/compression.rs
```diff
@@ -13,12 +14,14 @@ use std::ops::Neg;
...
-pub fn check_point_is_in_subgroup(point: &G1Point) -> bool {
-    let inf = G1Point::neutral_element();
-    let aux_point = point.operate_with_self(MODULUS);
+pub fn check_point_is_in_subgroup<SW: IsShortWeierstrass>(
+    point: &ShortWeierstrassProjectivePoint<SW>,
+) -> bool {
+    let inf = ShortWeierstrassProjectivePoint::<SW>::neutral_element();
+    let aux_point = point.operate_with_self(SUBGROUP_ORDER);
     inf == aux_point
 }
```
MauroToscano, line 26: It would be better for this to be in the pairing.rs file

MauroToscano (changes requested):
It doesn't look right. KZG is defined for any G1, G2. But the check is working over the BLS12 381. It also doesn't look right that `chek_point_is_in_subgroup` works over very shortweierstrass point, but actually, it's only working over the bls 381. The compiler is being tricked into thinking that the function is more generic than what it is

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/compression.rs
```diff
@@ -13,12 +14,14 @@ use std::ops::Neg;
...
 
-pub fn check_point_is_in_subgroup(point: &G1Point) -> bool {
-    let inf = G1Point::neutral_element();
-    let aux_point = point.operate_with_self(MODULUS);
+pub fn check_point_is_in_subgroup<SW: IsShortWeierstrass>(
+    point: &ShortWeierstrassProjectivePoint<SW>,
+) -> bool {
+    let inf = ShortWeierstrassProjectivePoint::<SW>::neutral_element();
+    let aux_point = point.operate_with_self(SUBGROUP_ORDER);
     inf == aux_point
```
schouhy, line 25: Here you can use a is a specific optimized method to check whether a point is the neutral element ([here](https://github.com/lambdaclass/lambdaworks/blob/21cc017bba911e0ada0822b291a180c393616c3b/math/src/elliptic_curve/short_weierstrass/point.rs#L83C18-L86)). In the case of Short Weierstrass, a point on the curve is the neutral element if and only if its Z coordinate is zero. So there's no need to compare all three coordinates.

schouhy (changes requested):

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/compression.rs
PatStiles: Thank you!

PatStiles:
@MauroToscano @schouhy Given your feedback I elected to add the subgroup check as a trait function for IsGroup. This may lead to deduplication if we have multiple coordinates system. Let me know your thoughts!

math/src/cyclic_group.rs
```diff
@@ -11,6 +11,11 @@ pub trait IsGroup: Clone + PartialEq + Eq {
         self == &Self::neutral_element()
     }
 
+    fn is_in_subgroup<T: IsUnsignedInteger>(&self, order: T) -> bool {
+        let aux_point = self.operate_with_self(order);
+        aux_point.is_neutral_element()
+    }
+
```
schouhy, line 18:
```suggestion
    /// This method checks whether the element belongs to the `order`-torsion subgroup.
    /// That is, the subgroup formed by all elements `P` such that (in additive notation) `order * P` is the neutral element.
    fn is_in_subgroup<T: IsUnsignedInteger>(&self, order: T) -> bool {
        self.operate_with_self(order).is_neutral_element()
    }
```

schouhy (changes requested):
Good work! Looks good. I left a suggestion and a request to remove a comment.

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/pairing.rs
```diff
@@ -25,18 +29,23 @@ impl IsPairing for BLS12381AtePairing {
     type OutputField = Degree12ExtensionField;
 
     /// Compute the product of the ate pairings for a list of point pairs.
+    //NOTE: Should we switch this to an result?
+    //Should we continue and skip elements if it doesn't work out
```
schouhy, line 33: A `Result` is good. Let's remove this comment

PatStiles:
@schouhy @MauroToscano should be good to go.

RajeshRk18:
@schouhy @MauroToscano I have implemented subgroup checks using endomorphism https://github.com/lambdaclass/lambdaworks/issues/543#issuecomment-1729937227

Should I make a separate PR titled optimized subgroup checks?

PatStiles:
@MauroToscano @schouhy I've started to work on adding Jacobian Coordinates so we can add the endomorphism used in [gnark](https://github.com/ConsenSys/gnark-crypto/blob/ed8216091919174f8d578ef1f023ebcb649df7fa/ecc/bls12-381/g1.go#L36C9-L36C9) Should I continue with this pr or open a new one?

schouhy:
@PatStiles is gnark's algorithm is strictly dependant on jacobian coordinates? That seems odd. I would expect it to be independent of the inner representation of the point. I'll check it out if I have time. But it would be useful if you could explain why we need jacobian coordinates for that.

MauroToscano:
New PR, keep this one naive and small

PatStiles:
@schouhy I haven't dived very deep into the implementation yet but will do so this afternoon. I noticed first when going over the gnark implementation the addition and mul operations differ. I investigating the alg reference https://hyperelliptic.org/EFD/g1p/auto-shortw-jacobian-3.html#addition-add-2007-bl are the slides found here used jacobian coordinates https://cr.yp.to/talks.html#2001.10.29. (pg 17). 

One thing to note is this could be a semantic change only as gnark's implementation has both [jacobian](https://github.com/ConsenSys/gnark-crypto/blob/ed8216091919174f8d578ef1f023ebcb649df7fa/ecc/bls12-381/g1.go#L36) and [extended jacobian](https://github.com/ConsenSys/gnark-crypto/blob/ed8216091919174f8d578ef1f023ebcb649df7fa/ecc/bls12-381/g1.go#L41) coordinates which appear to be whats normally referenced as Jacobian. 

Looking at there FromAffine() I think it is [projective](https://github.com/ConsenSys/gnark-crypto/blob/ed8216091919174f8d578ef1f023ebcb649df7fa/ecc/bls12-381/g1.go#L342)

Let me know your thoughts!

MauroToscano (approved):

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/pairing.rs
```diff
@@ -27,20 +32,90 @@ impl IsPairing for BLS12381AtePairing {
...
-        final_exponentiation(&result)
+        Ok(final_exponentiation(&result))
+    }
+}
+
+pub const CUBE_ROOT_OF_UNITY_G1: BLS12381FieldElement = FieldElement::from_hex_unchecked(
+    "5f19672fdf76ce51ba69c6076a0f77eaddb3a93be6f89688de17d813620a00022e01fffffffefffe",
+);
+
+pub const ENDO_U: BLS12381TwistCurveFieldElement =
```
schouhy, line 55: Please add docs about what these constant represent and add references to their sources. Also, prefer more declarative variable names for code readability: e.g. `TWIST_ENDOMORPHISM_U`

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/pairing.rs
```diff
@@ -27,20 +32,90 @@ impl IsPairing for BLS12381AtePairing {
...
+]);
+
+pub const ENDO_V: BLS12381TwistCurveFieldElement =
+BLS12381TwistCurveFieldElement::const_from_raw([
+    FieldElement::from_hex_unchecked("135203e60180a68ee2e9c448d77a2cd91c3dedd930b1cf60ef396489f61eb45e304466cf3e67fa0af1ee7b04121bdea2"),
+    FieldElement::from_hex_unchecked("6af0e0437ff400b6831e36d6bd17ffe48395dabc2d3435e77f76e17009241c5ee67992f72ec05f4c81084fbede3cc09")
+]);
+
+impl ShortWeierstrassProjectivePoint<BLS12381Curve> {
+    /// Returns phi(p) where `phi: (x,y)->(ux,y)` and `u` is the Cube Root of Unity in `G_1`
```
schouhy, line 68: roots of unity live in the prime field, not in the group.
```suggestion
    /// Returns phi(p) where `phi: (x,y)->(ux,y)` and `u` is the Cube Root of Unity in the base prime field
```

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/pairing.rs
```diff
@@ -27,20 +32,90 @@ impl IsPairing for BLS12381AtePairing {
...
+        Self::new([a.clone(), -b])
     }
 }
 
-/// This is equal to the frobenius trace of the BLS12 381 curve minus one.
+impl ShortWeierstrassProjectivePoint<BLS12381TwistCurve> {
+    fn is_on_curve(&self) -> bool {
+        let [x, y, z] = self.coordinates();
+        y.square() * z == (x.square() * x) + (z.square() * z * BLS12381TwistCurve::b())
+    }
```
schouhy, line 96: An instance of `ShortWeierstrassProjectivePoint<BLS12381TwistCurve>` should always satisfy the curve equation. If not, the constructor is failing. No need to check this.
And in any case, the curve has already a `definning_equation` method that implements this

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/pairing.rs
```diff
@@ -27,20 +32,90 @@ impl IsPairing for BLS12381AtePairing {
...
+
+    // psi(p) = u o frob o u**-1 where u:E'->E iso from the twist to E
```
schouhy, line 98:
```suggestion
    // psi(p) = u o frob o u**-1, where u is the isomorphism u:E'(𝔽ₚ₆) −> E(𝔽ₚ₁₂) from the twist to E
```

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/pairing.rs
```diff
@@ -27,20 +32,90 @@ impl IsPairing for BLS12381AtePairing {
...
+    fn psi(&self) -> Self {
+        let [x, y, z] = self.coordinates();
+        Self::new([
+            x.conjugate() * ENDO_U,
+            y.conjugate() * ENDO_V,
+            z.conjugate(),
+        ])
+    }
```
schouhy, line 106: This morphism should have minimal polynomial X^2 - tX + q (https://eprint.iacr.org/2022/352.pdf 4.2 equation (7)). Please add a test that checks this in a few points.
Also, please add a reference to the source of the formula.

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/pairing.rs
```diff
@@ -27,20 +32,90 @@ impl IsPairing for BLS12381AtePairing {
...
+
+    // https://eprint.iacr.org/2019/814.pdf, 3.1
+    // z*psi^3 - psi^2 + 1 = 0 <==> -z*(psi o phi) + phi + 1 = 0
+    pub fn is_in_subgroup(&self) -> bool {
+        let res = self
+            .psi()
+            .operate_with(&self.operate_with_self(MILLER_LOOP_CONSTANT));
+        res.is_on_curve() && res.is_neutral_element()
+    }
+}
```
schouhy, line 116: There's a simpler check https://eprint.iacr.org/2022/352.pdf 4.2. Just check psi(P) = zP.
And there's no need to check that te point is on the curve as it is a result of applying group opeartions to `self` that is already a point on the curve.

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/pairing.rs
```diff
@@ -27,20 +32,90 @@ impl IsPairing for BLS12381AtePairing {
...
+            .neg()
+            == self.phi()
+    }
+}
+
+impl FieldElement<Degree2ExtensionField> {
+    //TODO: Expose this function in extensions/qaudratic.rs
+    pub fn conjugate(&self) -> Self {
+        let [a, b] = self.value();
+        Self::new([a.clone(), -b])
```
schouhy, line 88: `Degree2ExtensionField` is a custom implementation that doesn't use the generic one in `extensions/quadratic.rs`.
Please move this implementation to `field_extension.rs` where `Degree2ExtensionField` is defined

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/pairing.rs
```diff
@@ -260,20 +335,78 @@ mod tests {
...
+    fn generator_g2_is_on_curve() {
+        let g = BLS12381TwistCurve::generator();
+        assert!(g.is_on_curve())
+    }
+
+    #[test]
+    fn arbitrary_g2_point_is_on_curve() {
+        let p = BLS12381TwistCurve::generator().operate_with_self(20_u64);
+        assert!(p.is_on_curve())
+    }
```
schouhy, line 402: These shouldn't be unit tests of pairing. Please move them to `curve.rs` and use the `defining_equation` method instead of `is_on_curve`

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/pairing.rs
```diff
@@ -260,20 +335,78 @@ mod tests {
...
+    fn generator_g2_is_in_subgroup() {
+        let g = BLS12381TwistCurve::generator();
+        assert!(g.is_in_subgroup())
+    }
+
+    #[test]
+    fn arbitrary_g2_point_is_in_subgroup() {
+        let p = BLS12381TwistCurve::generator().operate_with_self(20_u64);
+        assert!(p.is_in_subgroup())
+    }
```
schouhy, line 390: Please add tests of points that are in the curve but not in the subgroup, where the `p.is_in_subgroup` returns false.

schouhy (changes requested):
Hi! Thanks for your work!
I left a few requests. Let me know what you think.

Also, I would move all the code relevant to subgrup membership to the `curve.rs` file instead of `pairing.rs`.

PR #561, opened 2023-09-19
Update readmes with the info of the old repos

MauroToscano:
# Update readmes with the info of the old repos

- Created a Readme, based in the old ones, for Cairo prover and Stark Prover. 
- Updated main readme with a link to the documentation and added there references there. 
- Added a Makefile to the cairo prover so the CLI can be run as it was run before
- Added missing make clean to main Makefile
- Add link to the provers in the main README so they are easier to access

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -24,6 +24,9 @@ jobs:
           toolchain: stable
           targets: wasm32-unknown-unknown
 
+      - name: Set up cargo cache
+        uses: Swatinem/rust-cache@v2
+
       - name: Run cargo check
         run: cargo check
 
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yaml
```diff
@@ -24,6 +24,9 @@ jobs:
           toolchain: stable
           targets: wasm32-unknown-unknown
 
+      - name: Set up cargo cache
+        uses: Swatinem/rust-cache@v2
+
```
MauroToscano, line 29:
```suggestion
```

ajgara (approved):

schouhy (approved):

diegokingston (approved):

PR #563, opened 2023-09-19
Stark: Remove degree adjustment from composition poly

schouhy:
# Remove degree adjustment from composition poly

## Description
This PR removes the degree adjustment from the composition polynomial for compatibility with the stone prover.


ajgara (approved):

MauroToscano (approved):

MauroToscano (changes requested):
We should we update documentation @schouhy

diegokingston (approved):

MauroToscano (approved):

Issue #540, opened 2023-09-04
Fast Mersennes Prime Field 


MauroToscano:
- [ ] Implement the 32 bit Mersennes Prime Field, F = 2^31 - 1, with an efficient backend
- [ ] Implement the degree 4 field extension
- [ ] Implement mixed radix FFT/NTT for Mersennes Prime Field* (A compatible API with current FFT/NTT should be discussed)

It's recommended for new collaborators to only do the first task, since the code has no examples of field extensions for FRI

The last task needs discussion on how would be the best approach to implement it, API wise. 

For references, see:

[Reed-solomon over circle fields](https://eprint.iacr.org/2023/824.pdf)
[Plonky 3 implementation](https://github.com/Plonky3/Plonky3/blob/main/mersenne-31/src/lib.rs)

iammadab:
Will take this

PR #565, opened 2023-09-20
feat(field): add mersenne31 prime field

PatStiles:
# Adds Mersenne31 Prime Field

## Description

Starts #540 

Adds an implementation of the Mersenne31 prime field following the implementation in Plonky 3. 

- [x] New feature

## Checklist
- [x] Linked to Github Issue
- [x] Unit tests added


PatStiles:
@MauroToscano @schouhy Bench results for the Mersenne31 backend 
Screenshot 2023-09-22 at 00 18 26

MauroToscano:
> @MauroToscano @schouhy Bench results for the Mersenne31 backend in comparison to a U64 montgomery backend Screenshot 2023-09-22 at 00 18 26

The benchmark seems wrong, it's based in the old benchmarks of the Math library which are also not good. 

When using Criterion, the compiler may optimize when constants that are always the same. That's why you see one Pow taking 300 ps, which is highly suspicious. If you blackbox the paramater with "bench.iter(|| x.pow(black_box(*y)));", now the mersenne's backend is faster. Which is nice!

If you want to copy the ones of Math, you will have to fix this kind of issues. If you see some difference that big it may be the compiler optimizing something that it shouldn't be optimizing. 

The ones used here are more solid: https://github.com/lambdaclass/lambdaworks/tree/main/benches. We can improve the ones in math and fix this kind of issues.

PatStiles:
@MauroToscano Appreciate the feedback will get on it!

MauroToscano:
Looks good now

PatStiles:
@MauroToscano @schouhy I finished adding the field extensions and hooking into libfuzzer. One note is the lambdaworks [quadratic extension](https://github.com/lambdaclass/lambdaworks/blob/main/math/src/field/extensions/quadratic.rs#L90) differed slightly from the implementation of the inv() operation for the Mersenne31 fields complex extension in [plonky3](https://github.com/Plonky3/Plonky3/blob/main/mersenne-31/src/complex.rs#L249)

PR #567, opened 2023-09-20
Looped fibonacci 

MauroToscano:
# Add a looped fibonacci for benchmarks

This fibonacci is useful for comparing against other vms that are not validating the call stack, and benchmarks in general. It works as the other fibonaccis but uses less steps.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -24,6 +24,9 @@ jobs:
           toolchain: stable
           targets: wasm32-unknown-unknown
 
+      - name: Set up cargo cache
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -24,6 +24,9 @@ jobs:
           toolchain: stable
           targets: wasm32-unknown-unknown
 
+      - name: Set up cargo cache
+        uses: Swatinem/rust-cache@v2
+
```
MauroToscano, line 29:
```suggestion
```

ajgara (approved):

Oppen (approved):

PR #575, opened 2023-09-25
Change order of evaluation to avoid underflow error

mdvillagra:
# Change order of evaluation to avoid underflow error

## Description

When the number of limbs in an unsigned integer is 1, like in the babybear field, the line

```rust
let index = 2 * i - NUM_LIMBS + 1;
```

panicks. This is because all variables are `usize` and when `i=0` then `NUM_LIMBS` is subtracted and gives `-1`.

To solve the problem mentioned above is suffices to change the evaluation order.

## Type of change

- [x] Bug fix


diegokingston (approved):

pablodeymo (approved):
👍

PR #576, opened 2023-09-25
Make Babybear field FFT friendly

mdvillagra:
# Make Babybear field FFT friendly

## Description

Implemented `isFFT`trait to babybear field as requested in issue #539. Here the constants`TWO_ADIC_PRIMITVE_ROOT_OF_UNITY` and `TWO_ADICITY` were updated accordingly.
## Type of change

- [X] New feature

## Checklist
- [X] Linked to Github Issue
- [X] Unit tests added


math/src/field/fields/fft_friendly/babybear.rs
```diff
@@ -98,4 +116,20 @@ mod test_babybear_31_bytes_ops {
         let from_bytes = FieldElement::<Babybear31PrimeField>::from_bytes_be(&bytes).unwrap();
         assert_eq!(element, from_bytes);
     }
+
```
diegokingston, line 119: There should be a test using FFT to show that everything's working properly. Also, use cargo fmt to help with the code formatting

diegokingston (changes requested):

math/src/field/fields/fft_friendly/babybear.rs
mdvillagra: should the FFT test be added inside babybear.rs? or inside polynomial.rs like was done for `Stark252PrimeField`?

MauroToscano: You can add it inside babybear, but under the cfg[test], so it's only part of that module for testing purposes.

diegokingston (approved):

MauroToscano:
There's a bug with how the CUDA and Metal is used. It's assuming that if the feature is enabled, it can run with CUDA or Metal. As a quickfix, you can add this in the relevant tests`#[cfg(not(feature = "instruments"))]`. We should fix this in more general way, even if the feature is not enabled it could backoff to the CPU version, but it's not critical.

MauroToscano (approved):

Oppen (approved):

PR #577, opened 2023-09-25
Optimize build

MauroToscano:
# Optimize build

## Description

LTO = fat and Codegen 1 isn't improving performance, so we can remove it and compile faster. Neither is affecting the field benchmarks

## Type of change

- [x] Optimization


entropidelic (approved):

schouhy (approved):

MauroToscano:
This is having a penalty for EC operations, we should re add this for release

PR #580, opened 2023-09-27
Optimized subgroup checks

RajeshRk18:
# Optimized subgroup checks using Endomorphism

## Description

Existing subgroup check involves mutliplying a point $P$ by $r$ where $r$ is the prime order subgroup. This PR improves the performance of subgroup checks for $G_1$ and $G_2$ by using endomorphism. Theoritically, we will have 2x speedup for $G_1$ and 4x speedup for $G_2$. 

#559  implements Jacobian coordinates just for subgroup check which is not necessary.

## Type of change

- [ ] New feature
- [ ] Bug fix
- [x] Optimization

## Checklist
- [x] Linked to Github Issue
- [x] Unit tests added
  - [ ] Benchmarks added/run


math/src/elliptic_curve/short_weierstrass/curves/bls12_381/curve.rs
```diff
@@ -1,13 +1,25 @@
...
 use crate::elliptic_curve::traits::IsEllipticCurve;
+use crate::unsigned_integer::element::*;
 use crate::{
     elliptic_curve::short_weierstrass::traits::IsShortWeierstrass, field::element::FieldElement,
 };
 
 pub type BLS12381FieldElement = FieldElement<BLS12381PrimeField>;
 pub type BLS12381TwistCurveFieldElement = FieldElement<Degree2ExtensionField>;
 
+/// Cube Root of Unity in G1.
```
schouhy, line 13: Could you make this comment more clear?
```suggestion
/// Cube root of unity in the base field of BLS 12 381.
```

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/curve.rs
```diff
@@ -1,13 +1,25 @@
...
+pub const CUBE_ROOT_OF_UNITY_G1: U384 = U384::from_hex_unchecked(
+    "5f19672fdf76ce51ba69c6076a0f77eaddb3a93be6f89688de17d813620a00022e01fffffffefffe",
```
schouhy, line 15: Make this of type `FieldElement`. This will put it in Montgomery form once at compile time, instead of doing it at every call of `phi`.
```suggestion
pub const CUBE_ROOT_OF_UNITY_G1: FieldElement<BLS12381PrimeField> = FieldElement::from_hex_unchecked(
    "5f19672fdf76ce51ba69c6076a0f77eaddb3a93be6f89688de17d813620a00022e01fffffffefffe",
);
```

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/curve.rs
```diff
@@ -1,13 +1,25 @@
...
+);
+/// Cube Root of Unity in G2.
+pub const CUBE_ROOT_OF_UNITY_G2: U384 = U384::from_hex_unchecked("1a0111ea397fe699ec02408663d4de85aa0d857d89759ad4897d29650fb85f9b409427eb4f49fffd8bfd00000000aaac");
```
schouhy, line 18: Same comments as above (comment is unclear and make it `FieldElement<Degree2ExtensionField>`).

This is not used in your code. Its used though in `pairing.rs` ([here](https://github.com/lambdaclass/lambdaworks/blob/9825074e9269326bea93b788540b3bac7cde4056/math/src/elliptic_curve/short_weierstrass/curves/bls12_381/pairing.rs#L183)). So, I would leave it here and import it in `pairing.rs`

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/pairing.rs
```diff
@@ -27,16 +28,19 @@ impl IsPairing for BLS12381AtePairing {
     /// Compute the product of the ate pairings for a list of point pairs.
     fn compute_batch(
         pairs: &[(&Self::G1Point, &Self::G2Point)],
-    ) -> FieldElement<Self::OutputField> {
+    ) -> Result<FieldElement<Self::OutputField>, PairingError> {
         let mut result = FieldElement::one();
         for (p, q) in pairs {
+            if !(**p).is_in_subgroup() || !(**q).is_in_subgroup() {
```
schouhy, line 34:
```suggestion
            if !p.is_in_subgroup() || !q.is_in_subgroup() {
```

schouhy (commented):
Thanks for the PR! We'll be reviewing it in the next few days. Left some minor comments of a some things I saw.

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/twist.rs
```diff
@@ -46,7 +47,72 @@ impl IsShortWeierstrass for BLS12381TwistCurve {
...
+
+            let z_cubed_with_b =
+                <Self as IsEllipticCurve>::BaseField::mul(&z_cubed, Self::b().value());
+
+            <Self as IsEllipticCurve>::BaseField::add(&x_cubed, &z_cubed_with_b)
+        };
+
+        lhs == rhs
+    }
+}
```
schouhy, line 74: Something along this lines should do the work. Slower performancewise probably. But I wouldn't worry about that, since we didn't tackle optimizations of elliptic curve operations yet.
```suggestion
    pub fn is_on_curve(point: &<Self as IsEllipticCurve>::PointRepresentation) -> bool {
        if point.z() == &FieldElement::zero() {
            true
        } else {
            let point = point.to_affine();
            Self::defining_equation(&point.x(), &point.y()) == FieldElement::zero()
        }
    }
```
On the other hand. I saw that you use this method on a point that's already of type `ShortWeierstrassProjectivePoint<BLS12381TwistCurve>`. That shouldn't be necessary. If the current state of the code allows one to instantiate an element of that type that's not on the curve, then that's a bigger problem. If you suspect that's the case, please feel free to create an issue.

schouhy (commented):
Hi! I'm still going through it when I have time. I left some comments based on what I've reviewed so far. Let me know what you think.

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/field_extension.rs
```diff
@@ -244,6 +275,16 @@ mod tests {
...
 
+    #[test]
+    fn conjugate_works() {
+        let a: QuadraticExtensionFieldElement<MyQuadraticNonResidue> =
+            QuadraticExtensionFieldElement::zero();
+        let mut expected = a.conjugate();
+        expected = expected.conjugate();
+
+        assert_eq!(a, expected);
+    }
```
schouhy, line 286: This test is not testing that conjugate is properly implemented. It only tests that its square is the identity, which is also true for the identity function. I mean. `conjugate` could be implemented as the identity and this test would pass. 
But besides that. Why are you adding this test? This is functionality of generic quadratic extension fields and its already tested [here](https://github.com/lambdaclass/lambdaworks/blob/314fafcf44e1764dafa748f82d6df045f309cbc6/math/src/field/extensions/quadratic.rs#L250-L255).
Please remove it and also remove `MyQuadraticNonResidue` which is only used for this I guess.

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/field_extension.rs
```diff
@@ -139,7 +158,19 @@ impl ByteConversion for FieldElement<Degree2ExtensionField> {
...
+#[derive(Debug, Clone)]
+pub struct MyQuadraticNonResidue;
+
+impl HasQuadraticNonResidue for MyQuadraticNonResidue {
+    type BaseField = BLS12381PrimeField;
+
+    fn residue() -> FieldElement<Self::BaseField> {
+        -FieldElement::one()
+    }
+}
```
schouhy, line 172: Code for testing purposes should go in the `tests` module
```suggestion
```

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/field_extension.rs
```diff
@@ -105,6 +107,23 @@ impl IsField for Degree2ExtensionField {
...
 }
 
+impl Degree2ExtensionField {
+    /// Returns the frobenius map which is always conjugate of a field element.
+    pub fn frobenius_map(elem: &<Self as IsField>::BaseType) -> <Self as IsField>::BaseType {
+        let fe: FieldElement<QuadraticExtensionField<MyQuadraticNonResidue>> =
+            QuadraticExtensionFieldElement::new([elem[0].clone(), elem[1].clone()]);
+        let frob_map = fe.conjugate();
+
+        [frob_map.value()[0].clone(), frob_map.value()[1].clone()]
```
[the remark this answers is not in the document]
RajeshRk18: I saw `FieldElement<QuadraticExtensionField>` has conjugate fn. But it constraints to type that implement `HasQuadraticNonResidue`. So, I implemented it. If we can conjugate like you suggested, then why the `FieldElement<QuadraticExtensionField>` requires `HasQuadraticNonResidue` trait to be implemented to access conjugate fn?

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/curve.rs
```diff
@@ -8,6 +9,25 @@ use crate::{
...
+
+pub const CUBE_ROOT_OF_UNITY_G2_SQUARE: BLS12381TwistCurveFieldElement =
+    BLS12381TwistCurveFieldElement::const_from_raw([
+        FieldElement::<BLS12381PrimeField>::from_hex_unchecked(
+            "5f19672fdf76ce51ba69c6076a0f77eaddb3a93be6f89688de17d813620a00022e01fffffffefffe",
+        ),
+        FieldElement::<BLS12381PrimeField>::from_hex_unchecked("0"),
+    ]);
+
+/// Seed value for BLS-12 381 curve.
```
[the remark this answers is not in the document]
RajeshRk18: I got the value from [here](https://hackmd.io/@yelhousni/bls12_subgroup_check). There it is mentioned as seed.

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/twist.rs
```diff
@@ -46,7 +47,72 @@ impl IsShortWeierstrass for BLS12381TwistCurve {
...
+    }
+
+    /// Returns true if the point is in the prime subgroup `G_2` of order `r`.
+    /// Makes use of endomorphism for efficient point multiplication.
+    pub fn is_in_subgroup(&self) -> bool {
+        let seed_times_p = self.operate_with_self(SEED);
+        let psi_plus_seed_times_p = self.psi().operate_with(&seed_times_p);
+
+        Degree2ExtensionField::is_zero(psi_plus_seed_times_p.z().value())
+            && BLS12381TwistCurve::is_on_curve(&psi_plus_seed_times_p)
```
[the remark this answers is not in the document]
RajeshRk18: I saw that's how its implemented in [Gnark](https://github.com/Consensys/gnark-crypto/blob/ace5318b726c7185249bc9c92e1cadae701c1654/ecc/bls12-381/g2.go#L405). I should have listened to it that both checks for same thing.

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/field_extension.rs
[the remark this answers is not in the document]
schouhy: `QuadraticExtensionField` is a generic implementation of a quadratic extension for general non quadratic residues. The struct `Degree2Extension` is a custom impementation just for BLS12 381 to leverage some optimizations coming from the fact that the quadratic non residue is -1. For example, multiplication formulas for quadratic extensions in general involve multiplying by the quadratic non residue. When the residue is -1, there are aimpler formulas and one can avoid multiplying by 1 or -1. That's why for bls 12 381 we don't use the generic quadratic extension implementation.

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/twist.rs
RajeshRk18: Can you explain it more?

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/field_extension.rs
schouhy: Oh I see now that you were using it for the conjugate function. Not only for testing

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/field_extension.rs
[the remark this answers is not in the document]
schouhy: This should be explained in the docs of `Degree2Extension`. Sorry for that. I bet its confusing

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/twist.rs
schouhy: `ShortWeierstrassProjectivePoint<BLS12381TwistCurve>` is supposed to model a point on the twist. If there is an instance of it for which the defining equation of the twist does not hold, then there's a constructor failing to raise a creation error.

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/field_extension.rs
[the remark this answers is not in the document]
RajeshRk18: > `QuadraticExtensionField` is a generic implementation of a quadratic extension for general non quadratic residues. The struct `Degree2Extension` is an independent custom impementation just for BLS12 381 to leverage some optimizations coming from the fact that the quadratic non residue is -1. For example, multiplication formulas for quadratic extensions in general involve multiplying by the quadratic non residue. When the residue is -1, there are aimpler formulas and one can avoid multiplying by 1 or -1. That's why for bls 12 381 we don't use the generic quadratic extension implementation.

Thanks a lot for this explanation! Makes a lot more sense now. This is the true learning.

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/twist.rs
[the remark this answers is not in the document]
schouhy: Actually it doesn't really have to do with `z` equal to zero implying being equal to the point at infinity (since that already assumes that the point is on the curve). We should be able to rely on the fact that points of type `ShortWeierstrassProjectivePoint<BLS12381TwistCurve>` are on the curve. So a method such as `is_on_curve` should be redundant for instances of that type.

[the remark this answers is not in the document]
RajeshRk18: Yeah, it is because a point which may not be on curve will have z-coordinate set at 0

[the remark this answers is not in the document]
RajeshRk18: Maybe, `ShortWeierstrassProjectivePoint` can be made public at the crate level and add a builder fn that checks whether the point is on curve. Will that work?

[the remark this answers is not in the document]
schouhy: Yes. Prolly we only need the two builder functions we already have: [`create_point_from_affine`](https://github.com/lambdaclass/lambdaworks/blob/314fafcf44e1764dafa748f82d6df045f309cbc6/math/src/elliptic_curve/traits.rs#L26-L32) and [`deserialize` ](https://github.com/lambdaclass/lambdaworks/blob/314fafcf44e1764dafa748f82d6df045f309cbc6/math/src/elliptic_curve/short_weierstrass/point.rs#L213-L250). Both are checking the defining equation already.
But the attribute of the projective point is set to [public](https://github.com/lambdaclass/lambdaworks/blob/314fafcf44e1764dafa748f82d6df045f309cbc6/math/src/elliptic_curve/short_weierstrass/point.rs#L15C64-L15C86). So probably one could instantiate a mutable point from outside this crate and later change its coordinates. That would brake the "is on curve" invariant. That's a problem.

You can assume in this PR that any instance of type `ShortWeierstrassProjectivePoint<E>` is a point that's on the curve `E`. Let's discuss the other issue somewhere else. Feel free to create an issue for it (if you can provide an example and a candidate solution, even better!).

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/twist.rs
```diff
@@ -109,8 +95,7 @@ impl ShortWeierstrassProjectivePoint<BLS12381TwistCurve> {
         let seed_times_p = self.operate_with_self(SEED);
         let psi_plus_seed_times_p = self.psi().operate_with(&seed_times_p);
 
-        Degree2ExtensionField::is_zero(psi_plus_seed_times_p.z().value())
-            && BLS12381TwistCurve::is_on_curve(&psi_plus_seed_times_p)
+        psi_plus_seed_times_p.z() == &FieldElement::zero()
```
schouhy, line 98: points have a `is_neutral_element` already implemented with this logic. You can use that here too.

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/curve.rs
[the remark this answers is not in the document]
schouhy: Ah I see. For the sake of clarity, let's cite Construction 6.6 of the taxonomy of elliptic curves [paper](https://theory.stanford.edu/~dfreeman/papers/taxonomy.pdf) and mention that `SEED` is $-x$ where $x$ is the parameter of the Construction 6.6 for $k=12$. And add the formulas for the subgroup order and the field size in terms of $x$.
Check out this [site](https://fkunstner.github.io/latex-to-utf8/) to convert latex to ut8 and produce more legible formulas in code comments, (like x⁴ − x² + 1 instead of x^4 - x^2 + 1)

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/compression.rs
```diff
@@ -112,13 +104,13 @@ mod tests {
     fn test_zero_point() {
```
schouhy, line 104: Please rename this test

```suggestion
    fn test_point_not_in_subgroup() {
```

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/twist.rs
```diff
@@ -46,7 +47,57 @@ impl IsShortWeierstrass for BLS12381TwistCurve {
...
+        ])
+    }
+
+    /// Returns true if the point is in the prime subgroup `G_2` of order `r`.
+    /// Makes use of endomorphism for efficient point multiplication.
+    pub fn is_in_subgroup(&self) -> bool {
+        let seed_times_p = self.operate_with_self(SEED);
+        let psi_plus_seed_times_p = self.psi().operate_with(&seed_times_p);
+
+        BLS12381TwistCurve::is_on_curve(&psi_plus_seed_times_p)
```
schouhy, line 98: This is not checking that $\psi(P) - uP$ is the neutral element. You probably mean something different here.

```suggestion
        psi_plus_seed_times_p.is_neutral_element()
```

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/twist.rs
```diff
@@ -154,4 +205,16 @@ mod tests {
...
+    fn check_generator_g2_in_subgroup() {
+        let gen = BLS12381TwistCurve::generator();
+        assert!(gen.is_in_subgroup());
+    }
+
+    #[test]
+    fn check_arbitrary_g2_point_in_subgroup() {
+        let arb_point = BLS12381TwistCurve::generator().operate_with_self(420_u32);
+        assert!(arb_point.is_in_subgroup());
+    }
```
schouhy, line 219: Please add a test that checks that `is_in_subgroup` returns false for a point not in $G_2$.

math/src/elliptic_curve/short_weierstrass/curves/bls12_381/twist.rs
```diff
@@ -46,7 +47,57 @@ impl IsShortWeierstrass for BLS12381TwistCurve {
...
+        } else {
+            let point = point.to_affine();
+            Self::defining_equation(&point.x(), &point.y()) == FieldElement::zero()
+        }
+    }
+}
+
 impl ShortWeierstrassProjectivePoint<BLS12381TwistCurve> {
+    /// Returns the "Untwist-Frobenius-Twist" endomorphism
+    pub fn psi(&self) -> Self {
```
schouhy, line 63: Please add a test that checks that this endomorphism has minimal polynomial $X^2 - tX + q$. Maybe checking that on a few random points.

schouhy (changes requested):
Hi, thanks for incorporating all the previous changes!

I found an error on the subgroup check for G2. You prolly removed the wrong line in the previous version. Anyway, I left a few requests that aim to catch that kind of mistakes.

MauroToscano:
This is stale

PR #584, opened 2023-09-29
Crypto: Remove repeated code in Merkle tree backends

schouhy:
# Remove repeated code in Merkle tree backends

## Description

This PR leverages const generics to remove repeated code in `crypto/src/merkle_tree/backends`. We previously had two separate implementations for 256 and 512 bit in both batched and single element backends. This PR merges both into a single implementation for batch and a single implementation for single elements.

## Type of change

- [x] Refactor


diegokingston (approved):

ajgara (approved):

MauroToscano (approved):
Maybe we can find a better name for singlebackend ?

schouhy:
its indeed an ugly name. Other options I considered are: 

1. `FieldElementBackend` instead of `SingleBackend` and `FieldElementVectorBackend` instead of `BatchBackend`, referring to the data they work with.
2. Just `Backend` instead of `SingleBackend`
3. Just `Backend` for both of them and use them as `batch::Backend` and `single::Backend`

MauroToscano:
I think 1 Is nice. Also FEVector is more descriptive than Batch. We can change that change later though, I don't mind

Issue #592, opened 2023-10-05
Pedersen Hash


MauroToscano:
Tasks:

- [ ] Implement [Starknet Crypto equivalent ](https://docs.starkware.co/starkex/crypto/pedersen-hash-function.html) implementation
- [ ] Generalize over the elliptic curve, retain a predefined one with the Starknet curve

PR #597, opened 2023-10-09
stark curve, pedersen hash, precalculated points

irfanbozkurt:
# Pedersen Hash

## Description

Adds STARK curve, implements Pedersen hash and tests with values from [starknet-rs](https://github.com/xJonathanLEI/starknet-rs/), as requested in the first 2 checkpoints of https://github.com/lambdaclass/lambdaworks/issues/592.

## Type of change

- [x] New feature

## Checklist
- [x] Linked to Github Issue
- [x] Unit tests added


crypto/src/hash/pedersen/parameters.rs
```diff
@@ -0,0 +1,72 @@
...
+
+impl Default for PedersenParameters<StarkCurve> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl PedersenParameters<StarkCurve> {
+    pub fn new() -> Self {
+        let field_elements_csv = include_str!("stark/points.csv");
```
MauroToscano, line 32: It doesn't make much sense to have points in a CSV, just make a Rust file and have an array of FieldElement, or worst case, strings.

MauroToscano (commented):
We should add to references Starknet RS in the Readme

crypto/src/hash/pedersen/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+impl Pedersen<StarkCurve> {
+    pub fn new_with_params(params: PedersenParameters<StarkCurve>) -> Self {
+        Self { params }
+    }
+
+    /// Taken from Jonathan Lei's starknet-rs
+    /// https://github.com/xJonathanLEI/starknet-rs/blob/4ab2f36872435ce57b1d8f55856702a6a30f270a/starknet-crypto/src/pedersen_hash.rs
+    pub fn hash(
```
MauroToscano, line 33: This is great to have the reference, but it's not a good doc for the function. Here it should explain what does it do.

crypto/src/hash/pedersen/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+        element::FieldElement, fields::fft_friendly::stark_252_prime_field::Stark252PrimeField,
+    },
+};
+
+pub mod parameters;
+use self::parameters::PedersenParameters;
+
+pub struct Pedersen<EC: IsShortWeierstrass> {
+    params: PedersenParameters<EC>,
+}
```
MauroToscano, line 17: The generic when it only works for one curve seems a bit premature. But it's ok.

MauroToscano (approved):

diegokingston (approved):

crypto/src/hash/pedersen/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+                bits.chunks(self.params.curve_const_bits)
+                    .enumerate()
+                    .for_each(|(i, v)| {
+                        let offset = bools_to_usize_le(v);
+                        if offset > 0 {
+                            // Table lookup at 'offset-1' in table for chunk 'i'
+                            *acc = acc.operate_with(&prep[i * self.params.table_size + offset - 1]);
+                        }
+                    });
+            };
```
entropidelic, line 51: This seems like it could be a separate private function instead of defining it here. Is there any particular reason to do it?

Also, lets add some documentation about what this function does. Is it just adding elliptic curve points? Maybe the name could be better too, since it is not very clear what it does.

entropidelic:
good work @irfanbozkurt, thanks! just made a little suggestion so that code is clearer

crypto/src/hash/pedersen/mod.rs
irfanbozkurt: Hey, thanks for the feedback.

The only reason was I tried to make as little changes as possible to the reference implementation. 

I now defined it as a private function as you suggested, and renamed it into "lookup_and_accumulate", as well as putting some explanations. I'm not sure if this is a great name but it sounds more explanatory.

entropidelic (approved):

PR #598, opened 2023-10-09
migrate cli to extern crate

juan518munoz:
# CAIRO PROVER CLI

## Description

Migration of cli functionallity to an external crate

## Type of change

- [x] Optimization

## Checklist
- [x] Linked to Github Issue


diegokingston (approved):

entropidelic:
The PR looks good! nice work.
I have some questions/suggestions:
* Now that we use clap for the CLI, should we continue using the Makefile CLI? Maybe it is ok to have both, but the interface using clap directly seems nicer to me. In this case, me should add to the docs how to use it directly, without the need of make.
* Maybe not too relevant right now, but in the future as the number of arguments for the CLI grows, we may want to specify them with names (for example, for specifying the program `--program <compiled_program>`, etc). This way the intent is clearer and the order of the arguments doesn't matter, making it a little bit friendlier for the user.

cc @MauroToscano

MauroToscano (approved):

entropidelic:
Lets merge this now and add CLI usage instructions in another PR

entropidelic (approved):

PR #599, opened 2023-10-10
Add Table data structure and refactor code

entropidelic:
A new `Table` data structure is added, which acts as a backend for tabular data structures used in the protocol implementation, namely:
* `TraceTable`
* `Frame`

Moreover, some quality of life changes were made to the code:
* Some variable names were changed into clearer ones
* Minor refactors were made to increase code legibility
* Documentation was added to functions, especially to `Table` and `TraceTable` methods
* Some sanity checks were added across the code


provers/stark/src/trace.rs
```diff
@@ -1,125 +1,142 @@
...
+    /// such as to sort them.
+    pub fn merge_columns(&self, column_indexes: &[usize]) -> Vec<FieldElement<F>> {
+        let mut data = Vec::with_capacity(self.n_rows() * column_indexes.len());
+        for row_index in 0..self.n_rows() {
+            for column in column_indexes {
+                data.push(self.table.data[row_index * self.n_cols() + column].clone());
+            }
+        }
+        data
+    }
```
MauroToscano, line 81: Can we work with an iterator and avoid cloning all the values ?

MauroToscano (approved):

provers/stark/src/trace.rs
MauroToscano: For example:

```
    let v1 = vec![1,2,3];
    let v2 = vec![11,22,33];
    
    let iter = v1.iter().chain(v2.iter());  
```

provers/stark/src/table.rs
```diff
@@ -0,0 +1,108 @@
...
+            return Self {
+                data: Vec::new(),
+                width,
+                height: 0,
+            };
+        }
+
+        // Check that the one-dimensional data makes sense to be interpreted as a 2D one.
+        debug_assert!(crate::debug::check_2d_array_integrity(data, width));
+        let height = data.len() / width;
```
MauroToscano, line 30: Should this be a debug_assert ? Shouldn't this raise an error ?

MauroToscano: Also, it's not clear what this function does. Maybe the name could be better

entropidelic: I agree with the debug assertion, for the moment I can use a normal `assert!()` until we have better error handling.

entropidelic: I renamed the function to `validate_2d_structure`, what do you think?

diegokingston (approved):

schouhy:
Awesome!
There's also an `EvaluationTable` [here](https://github.com/lambdaclass/lambdaworks/blob/table-refactor/provers/stark/src/constraints/evaluation_table.rs). It's only used to hold the evaluations of the composition polynomial and its method `compute_composition_poly` is just calling fft. We can probably remove it too.

provers/stark/src/table.rs
MauroToscano: Ok, we. can think of a better name later

schouhy (approved):

PR #618, opened 2023-10-20
Merkle cli

juan518munoz:
# Merkle tree CLI

## Description
image

1. Generate Merkle trees and serialize them into a `.json` file.
2. Create Proof for the nth element inside a Merkle tree.
3. Verify inclusion from a proof, root, index and element value.

⚠️ Currently both `1.` and `2.` read trees from `.csv` files. To have a better workflow it may be better to make `2.` use the trees (`.json` files) generated in the previous point.  ⚠️

## Type of change
- [x] New feature

## Checklist


examples/merkle-tree-cli/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+```bash
+cargo run --release generate-merkle-tree sample_tree.csv
+```
+
+
+**To generate proof for a Merkle Tree you can use:**
+
+```bash
+cargo run --release generate-merkle-tree <TREE_PATH> <POSITION>
```
entropidelic, line 27: this should call the command `generate-proof`

examples/merkle-tree-cli/README.md
```diff
@@ -0,0 +1,78 @@
...
+For example:
+
+```bash
+cargo run --release generate-merkle-tree sample_tree.csv 0
+```
+
+**To verify a proof you can use:**
+
+```bash
+cargo run --release generate-merkle-tree <ROOT_PATH> <INDEX> <PROOF_PATH> <LEAF_PATH>
```
entropidelic, line 39: this should call the `verify` command

examples/merkle-tree-cli/sample_tree.csv
```diff
@@ -0,0 +1,3 @@
+0x12345
+0x6789A
+0xBCDEF
```
entropidelic, line 3: I don't really know if this changes anything, but this is not a `csv` file.

examples/merkle-tree-cli/src/commands.rs
```diff
@@ -0,0 +1,37 @@
...
+#[command(author = "Lambdaworks", version, about)]
+pub struct MerkleArgs {
+    #[clap(subcommand)]
+    pub entity: MerkleEntity,
+}
+
+#[derive(Subcommand, Debug)]
+pub enum MerkleEntity {
+    #[clap(about = "Generate a merkle tree")]
+    GenerateMerkleTree(GenerateMerkleTreeArgs),
```
entropidelic, line 13: Just a personal opinion, maybe this could be just "GenerateTree", by context we already know that it is a merkle tree and the command becomes shorter :)

examples/merkle-tree-cli/src/main.rs
```diff
@@ -0,0 +1,97 @@
...
+
+fn generate_merkle_proof(tree_path: String, pos: usize) -> Result<(), io::Error> {
+    let values: Vec<FE> = load_tree_values(&tree_path)?;
+
+    let merkle_tree = MerkleTree::<Poseidon<BLS12381PrimeField>>::build(&values);
+
+    let Some(proof) = merkle_tree.get_proof_by_pos(pos) else {
+        return Err(io::Error::new(
+            io::ErrorKind::Other,
+            "Could not generate proof",
```
entropidelic, line 54: This error could be a little better, maybe informing the user that the position was not found in the tree

examples/merkle-tree-cli/src/main.rs
```diff
@@ -0,0 +1,97 @@
...
+use lambdaworks_math::{
+    elliptic_curve::short_weierstrass::curves::bls12_381::field_extension::BLS12381PrimeField,
+    field::element::FieldElement,
+};
+use std::{
+    fs::{self, File},
+    io::{self, BufWriter, Write},
+};
+
+type FE = FieldElement<BLS12381PrimeField>;
```
entropidelic, line 17: We should use the Starkfield instead

examples/merkle-tree-cli/README.md
entropidelic: should the file be the original `csv` or the `json` tree?

entropidelic: Oh, I see that was mentioned in the PR description, sorry

examples/merkle-tree-cli/README.md
entropidelic: I could not run this command successfully, I get the error 
```
Error while running command: Custom { kind: Other, error: "InvalidHexString" }
```

It should be explained how to generate the `root` and `leaf` files. 
I tried with a txt file with the plain numbers and I got the earlier error

examples/merkle-tree-cli/README.md
juan518munoz: > this should call the command `generate-proof`

Correct, modified.

> should the file be the original csv or the json tree?

Which file do you think is the better option?

examples/merkle-tree-cli/README.md
juan518munoz: > this should call the `verify-proof` command

Yes, changing it.

> I could not run this command successfully

I've added documentation on the README on how both input files should be.

examples/merkle-tree-cli/sample_tree.csv
juan518munoz: While it could be argued that it's a [valid csv](https://en.wikipedia.org/wiki/Comma-separated_values#:~:text=CSV%20is%20a%20delimited%20data%20format%20that%20has%20fields/columns%20separated%20by%20the%20comma%20character%20and%20records/rows%20terminated%20by%20newlines.) considering that each element is read as a row and not a column, I understand that a `csv` file that separetes elements by commas is the more standarized approach. I have made the changes necessary to make it work that way.

examples/merkle-tree-cli/src/commands.rs
juan518munoz: Yes, the other two commands also ommit the word _Merkle_. Changing it.

examples/merkle-tree-cli/src/main.rs
juan518munoz: This one is a bit complicated, as the function `get_proof_by_pos` calls inside `build_merkle_path`, which panics the given index is out of bounds. 

https://github.com/lambdaclass/lambdaworks/blob/merkle-cli/crypto/src/merkle_tree/merkle.rs#L37-L58

juan518munoz: Modified `get_proof_by_pos` inside `merkle.rs` so it can cath an index out of bounds and return an `Err` if that happens.

juan518munoz:
[c02a3ab](https://github.com/lambdaclass/lambdaworks/pull/618/commits/c02a3ab9a458d3becdd1a29f3a2e640522de9a2e) breaks some parts of the codebase that relied on `get_proof_by_pos`, working on a fix.

juan518munoz:
> ⚠️ [c02a3ab](https://github.com/lambdaclass/lambdaworks/pull/618/commits/c02a3ab9a458d3becdd1a29f3a2e640522de9a2e) breaks some parts of the codebase that relied on `get_proof_by_pos`, working on a fix.

Fixed.

examples/merkle-tree-cli/sample_tree.csv
entropidelic: thanks!

examples/merkle-tree-cli/README.md
entropidelic: I think the original csv seems a better option

juan518munoz: Leaving as is then.

entropidelic (approved):

MauroToscano (approved):

PR #620, opened 2023-10-20
fix: polish some CLI details

entropidelic:
Some minor details from the CLI and readme were corrected:
* when executing `compile-and-prove`, the proof was not being serialized
* the `run_all` command in the readme was not longer supported, replaced with `prove-and-verify` instead.
* replaced `compile_and_run_all` for `compile-prove-and-verify`
* refactor serialization of proof in the CLI into a function

### DIFF
--- a/provers/cairo/README.md
+++ b/provers/cairo/README.md
@@ -97,13 +97,13 @@ cargo run --release --features="cli" prove-and-verify cairo_programs/cairo0/fibo
 **To compile, proof, prove and verify at the same time you can use:**
 
 ```bash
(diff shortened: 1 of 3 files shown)


diegokingston (approved):

juan518munoz (approved):
with `run_all` renamed to `prove-and-verify`, maybe `compile-and-run-all` should be changed to `compile-prove-and-verify`

entropidelic:
@juan518munoz I agree, now the command is updated

diegokingston (approved):

MauroToscano (approved):

PR #634, opened 2023-10-27
add new cli command to prove traces

mb-dci:
Sparkling Water Bootcamp task 

Add support for the prover to generate proofs from memory.bin and trace.bin executions of Cairo programs.

+ prove-traces command added
+ integration test added


aniketpr01:
Current implementation has no builtin support, and range_check output is set to None.
We have only added one unit test of fibonacci_5 program as of now, if suggested, then will add for other programs as well. It may require to modify docker files to run with cairo-run entrypoint so that it will be able to generate the trace files when executed during make test, as in the current implementation we have just uploaded the trace files for fibonacci_5 program.

MauroToscano (approved):

MauroToscano:
Looking good

MauroToscano:
It would be cool to add this to the readme

MauroToscano (changes requested):
Add to the readme the new command and how to use it with cairo-run

provers/cairo/src/commands.rs
```diff
@@ -13,6 +13,8 @@ pub enum ProverEntity {
     Compile(CompileArgs),
     #[clap(about = "Generate a proof for a given compiled cairo program")]
     Prove(ProveArgs),
+    #[clap(about = "Generate a proof from a given trace of a cairo program execution")]
+    ProveTrace(ProveTraceArgs),
```
entropidelic, line 17: Let's make a little change of names; I think it is more accurate if the `Prove` command is renamed to something like "RunAndProve" (the command description should be changed too), and the new command you implemented should be the one named `Prove`.
This reflects better the intent of both commands IMO

provers/cairo/src/main.rs
```diff
@@ -218,6 +249,22 @@ fn main() {
 
             write_proof(proof, pub_inputs, args.proof_path);
         }
+        commands::ProverEntity::ProveTrace(args) => {
+            // verify input files
+            if !args.trace_bin_path.contains(".bin") && !args.memory_bin_path.contains(".bin") {
+                eprintln!("\n Please provide *.bin files for the memory and trace output files.\n");
+                return;
+            }
```
entropidelic, line 257: I think this check isn't really necessary, lets be more flexible on the files format. It really doesn't matter what the file extension is

entropidelic (approved):

entropidelic:
great work 🎉

diegokingston (approved):

MauroToscano (approved):

PR #650, opened 2023-10-31
Parallel trace compute

MauroToscano:
# Parallel trace compute

## Description

Add parallel trace compute, fibo 1000:

 735.929042ms  ->  672.014791ms 

## Type of change

- [x] Optimization

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,3 +21,12 @@ cairo-platinum-prover = { path = "./provers/cairo", version = "0.2.0" }
 lto = true
 codegen-units = 1
 opt-level = 3
+
+[profile.ci]
+inherits = "release"
+opt-level = 3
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -21,3 +21,12 @@ cairo-platinum-prover = { path = "./provers/cairo", version = "0.2.0" }
...
 opt-level = 3
+
+[profile.ci]
+inherits = "release"
+opt-level = 3
+lto = false
+
+# 131 s no linker o2
+# 110 llm o2 linker
+# 102 llm o1 linker
```
MauroToscano, line 32:
```suggestion
```

MauroToscano, line 32:
```suggestion
```

PR #653, opened 2023-10-31
sample ser/de for structs

juan518munoz:
# Proof serialization fix

⚠️ This code won't compile ⚠️

## Description

The aim of this pull request is to minimize the file size of `.proofs` caused by `Field` elements being stored as strings, code provided shows examples on how structures that use `FieldElements` need to implement `serde::Serialization` and `serde::Deserialization` need to accomodate to the fact that not all `FieldElements` can be serialized.

### Serde documentation:
[Struct serialization](https://serde.rs/impl-serialize.html#serializing-a-struct)
[Struct deserialization](https://serde.rs/deserialize-struct.html)

## Type of change

- [x] Optimization

If this approach is the one to be used, documentation on how to implement serialization and deserializaiton should be added.

### DIFF
--- a/math/src/field/element.rs
+++ b/math/src/field/element.rs
@@ -1,6 +1,9 @@
+use crate::elliptic_curve::short_weierstrass::curves::bls12_381::compression::check_point_is_in_subgroup;
 use crate::errors::CreationError;
 use crate::field::errors::FieldError;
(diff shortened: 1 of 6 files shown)


juan518munoz:
Closed in favour of #657

PR #657, opened 2023-11-01
Proof serialization fix alt

juan518munoz:
# Proof serialization - reduced size

## Description

The aim of this pr is to replace the current `cbor` serializer with `bincode`, which creates a serialized file of around half the original size.

Note that to make this implementation work, the `BaseType` type is now bound to the `ByteConversion` trait. Fields that previously didn't implement `ByteConversion` have the barebones implementations needed filled with `todo!()`, so a followup pr should be added with the proper implementations. 

## Type of change

- [x] Optimization

## Checklist
  - [ ] Benchmarks added/run

### DIFF
--- a/math/src/field/fields/montgomery_backed_prime_fields.rs
+++ b/math/src/field/fields/montgomery_backed_prime_fields.rs
@@ -470,7 +470,7 @@ mod tests_u384_prime_fields {
         let x = U384F23Element::from(11_u64);
         let x_serialized = serde_json::to_string(&x).unwrap();
         let x_deserialized: U384F23Element = serde_json::from_str(&x_serialized).unwrap();
-        assert_eq!(x_serialized, "{\"value\":\"0xb\"}");
+        // assert_eq!(x_serialized, "{\"value\":\"0xb\"}"); // serialization is no longer as hex string
         assert_eq!(x_deserialized, x);
     }
 
--- a/provers/cairo/Cargo.toml
+++ b/provers/cairo/Cargo.toml
@@ -40,7 +40,6 @@ rayon = { version = "1.8.0", optional = true }
 wasm-bindgen = { version = "0.2", optional = true }
 serde-wasm-bindgen = { version = "0.5", optional = true }
 web-sys = { version = "0.3.64", features = ['console'], optional = true }
-rmp-serde = "1.1.2"
 
 [dev-dependencies]
 hex = "0.4.3"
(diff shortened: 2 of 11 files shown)


math/src/field/fields/montgomery_backed_prime_fields.rs
```diff
@@ -470,7 +470,7 @@ mod tests_u384_prime_fields {
         let x = U384F23Element::from(11_u64);
         let x_serialized = serde_json::to_string(&x).unwrap();
         let x_deserialized: U384F23Element = serde_json::from_str(&x_serialized).unwrap();
-        assert_eq!(x_serialized, "{\"value\":\"0xb\"}");
+        // assert_eq!(x_serialized, "{\"value\":\"0xb\"}"); // serialization is no longer as hex string
```
entropidelic, line 473: should this be removed?

provers/cairo/src/main.rs
```diff
@@ -198,6 +201,7 @@ fn main() {
                 eprintln!("\nYou are trying to prove a non compiled Cairo program. Please compile it before sending it to the prover.\n");
                 return;
             }
+            println!("{:?}", args.program_path);
```
entropidelic, line 204: seems like a debug print. Should it be removed?

entropidelic (approved):

provers/cairo/src/main.rs
juan518munoz: Yes, deleting it.

math/src/field/fields/montgomery_backed_prime_fields.rs
juan518munoz: [It was previously serialized as a hex string](https://github.com/lambdaclass/lambdaworks/blob/cffa505af7c0c9f73d938577958be20e6b92676a/math/src/field/element.rs#L445), the way it's done now makes the serialized bytes not human readable.

image

I assume that the assert was only done to check that the readability was ok, as the next assertion done on the test already verifies that the (de)serealization is done correctly.

provers/cairo/Cargo.toml
```diff
@@ -40,6 +40,7 @@ rayon = { version = "1.8.0", optional = true }
 wasm-bindgen = { version = "0.2", optional = true }
 serde-wasm-bindgen = { version = "0.5", optional = true }
 web-sys = { version = "0.3.64", features = ['console'], optional = true }
+rmp-serde = "1.1.2"
```
MauroToscano, line 43: RMP is not currently used
```suggestion
```

MauroToscano (changes requested):

provers/cairo/Cargo.toml
juan518munoz: removed

provers/cairo/src/main.rs
MauroToscano, line 204:
```suggestion
```

MauroToscano (approved):

diegokingston (approved):

PR #658, opened 2023-11-01
docs: stone prover trace docs - layout plain

entropidelic:
* Add docs for the trace configuration used in the Stone prover
* Add representation of the trace built in the stone prover in csv format, for a simple fibonacci program.


diegokingston (approved):

MauroToscano (approved):

MauroToscano:
Maybe we should add this to the main docs

ajgara (approved):

PR #668, opened 2023-11-06
refactor: change `Frame` to `Table` struct in stark proof for OODS values

entropidelic:
* Changed the struct holding the OOD trace evaluations from `Frame` to `Table`
* The `Table` constructor now consumes the `data` used for construction. There is no current use case for preserving the `data` outside of the table creation. Moreover, this makes the code a little bit simpler


provers/stark/src/table.rs
```diff
@@ -15,7 +15,7 @@ pub struct Table<F: IsFFTField> {
 impl<F: IsFFTField> Table<F> {
     /// Crates a new Table instance from a one-dimensional array in row major order
     /// and the intended width of the table.
-    pub fn new(data: &[FieldElement<F>], width: usize) -> Self {
```
MauroToscano, line 18: Is this needed ?

entropidelic: I think it is better for the `data` vector to be consumed once the `Table` struct is initialized with it. I can't think of any use case where we want to preserve the `data` vector by itself, what do you think?

diegokingston (approved):

MauroToscano (approved):

ajgara (approved):

Issue #655, opened 2023-11-01
Improve performance of Plonk prover


diegokingston:
[Here](https://github.com/lambdaclass/lambdaworks/blob/cffa505af7c0c9f73d938577958be20e6b92676a/provers/plonk/src/prover.rs#L356) we compute the terms to get the evaluations of the z polynomial. We should change it so that we leverage the use of batch inverse, instead of doing one inversion per step.

PR #671, opened 2023-11-06
Improvements over plonk

irfanbozkurt:
## Description

- Satisfies change requested in https://github.com/lambdaclass/lambdaworks/issues/655
- Same domain (powers of omega) was being calculated 4 times. Reduces that to only once
- Introduces successors to some places instead of instantiating a vector and pushing

### DIFF
--- a/provers/plonk/src/prover.rs
+++ b/provers/plonk/src/prover.rs
@@ -341,33 +341,38 @@ where
     fn round_2(
         &self,
         witness: &Witness<F>,
-        common_preprocessed_input: &CommonPreprocessedInput<F>,
+        cpi: &CommonPreprocessedInput<F>,
         beta: FieldElement<F>,
         gamma: FieldElement<F>,
     ) -> Round2Result<F, CS::Commitment> {
-        let cpi = common_preprocessed_input;
-        let mut coefficients: Vec<FieldElement<F>> = vec![FieldElement::one()];
+        let mut z_evaluations: Vec<FieldElement<F>> = vec![FieldElement::one()];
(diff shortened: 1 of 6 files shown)


provers/plonk/src/prover.rs
```diff
@@ -341,33 +341,38 @@ where
...
         let (s1, s2, s3) = (&cpi.s1_lagrange, &cpi.s2_lagrange, &cpi.s3_lagrange);
 
         let k2 = &cpi.k1 * &cpi.k1;
 
         let lp = |w: &FieldElement<F>, eta: &FieldElement<F>| w + &beta * eta + &gamma;
 
+        let z_nums: Vec<_> = (0..&cpi.n - 1)
+            .map(|i| {
+                lp(&witness.a[i], &cpi.domain[i])
+                    * lp(&witness.b[i], &(&cpi.domain[i] * &cpi.k1))
```
diegokingston, line 358: Maybe this could be more efficient just by just iterating over cpi.domain zipped with witness.a, witness.b, witness.c

provers/plonk/src/test_utils/utils.rs
```diff
@@ -34,51 +34,55 @@ pub fn test_srs(n: usize) -> StructuredReferenceString<G1Point, G2Point> {
...
-    let mut result: Vec<FieldElement<F>> = vec![];
-    for index_column in 0..=2 {
-        for index_row in 0..n {
-            result.push(w.pow(index_row) * u.pow(index_column as u64));
-        }
-    }
-    result
+    let u_powers = [&FieldElement::one(), &u, &(u * u)];
+
+    (0..=2)
```
diegokingston, line 77: There is no need to start from zero if it is just multiplying with one

PR #676, opened 2023-11-09
Public inputs refactor

juan518munoz:
# Refactor Public Inputs

## Description

Modify the `MemorySegmentMap` field of `PublicInputs` to be a HashMap with values of type `Segment` instead of `Range`.

This pr also updates the cairo-vm crate to it's latest `main` revision.

## Type of change

- [x] Optimization


juan518munoz:
Initial implementation, the struct `Segment` replaces `Range<u64>` as the value of the `MemorySegmentMap : HashMap`.

`Range<u64>` is still used (and instiated like `n..m`) on some functions, for this reason `From` and `Into` has been implemented beetwen this types.

juan518munoz:
Should we implement `Iterator` for `Segment` in a [similar way `Range` does](https://doc.rust-lang.org/src/core/iter/range.rs.html#707)?

entropidelic:
>Should we implement Iterator for Segment in a [similar way Range does](https://doc.rust-lang.org/src/core/iter/range.rs.html#707)?

@juan518munoz I think it is not necessary right now, is there any benefit on doing so?

juan518munoz:
> > Should we implement Iterator for Segment in a [similar way Range does](https://doc.rust-lang.org/src/core/iter/range.rs.html#707)?
> 
> @juan518munoz I think it is not necessary right now, is there any benefit on doing so?

Not many, it would only save us from doing two `clones()` that are needed to cast into `Range`.

diegokingston (approved):

provers/cairo/src/runner/run.rs
```diff
@@ -122,22 +125,46 @@ pub fn run_program(
     let cairo_mem = CairoMemory::from_bytes_le(&memory_vec).unwrap();
     let register_states = RegisterStates::from_bytes_le(&trace_vec).unwrap();
 
-    let data_len = runner.get_program().data_len();
+    let codelen = runner.get_program().data_len();
+
+    let public_memory = (1..=codelen as u64)
+        .map(|i| (Felt252::from(i), *cairo_mem.get(&i).unwrap()))
+        .collect::<HashMap<Felt252, Felt252>>();
```
entropidelic, line 132: Now that we have the public inputs directly from the cairo vm, there is no need to ask for the codelen, and the public memory should be obtained from the public inputs, this was just a hacky solution until we had access to them.
The approach should be similar to the one you did with the memory segments

entropidelic: As a result of this, the `PublicInputs` field `codelen` could be removed, since it can be deduced with the length of the `Program` memory segment

entropidelic (changes requested):
Excellent work on this one, only requested a little but important detail for this refactor

provers/cairo/src/runner/run.rs
juan518munoz: The only function that makes use of the `codelen` field is `get_memory_holes`, should we remove it?

juan518munoz: > The approach should be similar to the one you did with the memory segments

I'm having trouble doing an iteration over the vm public memory, doing something like this:
```rs
    let mut public_memory: HashMap<Felt252, Felt252> = HashMap::new();
    vm_public_inputs.public_memory.iter().for_each(|e| {
        let v = e.value.clone().unwrap().to_str_radix(16);
        public_memory.insert(
            Felt252::from((e.address + e.page) as u64),
            Felt252::from_hex_unchecked(&v),
        );
    });
   ```
   
Breaks some of our tests
   
image

Maybe I'm missing something?

**Review by entropidelic — APPROVED** [review]

File: provers/cairo/src/air.rs [hunk 2]
```diff
@@ -160,12 +158,60 @@ pub const MEM_P_TRACE_OFFSET: usize = 17;
...
+        let begin_addr = begin_addr.try_into().unwrap();
+        let stop_ptr = stop_ptr.try_into().unwrap();
+
+        Self {
+            begin_addr,
+            stop_ptr,
+        }
+    }
+    pub fn segment_size(&self) -> Option<usize> {
+        self.stop_ptr.checked_sub(self.begin_addr - 1)
```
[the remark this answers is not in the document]
Oppen: Also, for the first address (code segment, offset `0`) this will panic: `self.begin_addr - 1` is negative, which is out of range for `usize`. Maybe you want to add `1` after the subtraction instead?

provers/cairo/src/air.rs
```diff
@@ -487,9 +529,7 @@ fn add_pub_memory_in_public_input_section(
     let mut a_aux = addresses.to_owned();
     let mut v_aux = values.to_owned();
 
-    let output_range = public_input.memory_segments.get(&MemorySegment::Output);
-
-    let pub_addrs = get_pub_memory_addrs(output_range, public_input);
+    let pub_addrs: Vec<Felt252> = public_input.public_memory.clone().into_keys().collect();
     let mut pub_addrs_iter = pub_addrs.iter();
```
Oppen, line 533: This looks like you just want to iterate the keys. Why copy the whole public memory?

PR #685, opened 2023-11-14
Add periodic columns

ajgara:
Adds support for periodic columns and a small Fibonacci example using a periodic column for testing purposes


provers/stark/src/traits.rs
```diff
@@ -122,4 +123,23 @@ pub trait AIR: Clone {
...
+    fn get_periodic_column_polynomials(&self) -> Vec<Polynomial<FieldElement<Self::Field>>> {
+        let mut result = Vec::new();
+        for periodic_column in self.get_periodic_column_values() {
+            let values: Vec<_> = periodic_column
+                .iter()
+                .cycle()
+                .take(self.trace_length())
+                .cloned()
+                .collect();
+            let poly = Polynomial::interpolate_fft(&values).unwrap();
```
diegokingston, line 140: Maybe we can make this more efficient by using the appropriate iterator, but we can leave it for optimization

diegokingston (approved):

MauroToscano (approved):

PR #688, opened 2023-11-18
Generalize Pedersen hash function over the elliptic curve

techiepriyansh:
This PR attempts a very basic generalization of the Pedersen Hash over the Elliptic Curve. It currently only abstracts out the elliptic curve from the [Starknet Crypto equivalent](https://docs.starkware.co/starkex/crypto/pedersen-hash-function.html) implementation.

From what I can tell, the current Starknet equivalent implementation uses a simplified version of the generalized Pedersen Hash described in the [Iden3 docs](https://iden3-docs.readthedocs.io/en/latest/iden3_repos/research/publications/zkproof-standards-workshop-2/pedersen-hash/pedersen.html). Please let me know if it instead needs to be implemented in its full generality.

Aims to address the 3rd checkpoint of #592.

### DIFF
--- a/benches/benches/point.rs
+++ b/benches/benches/point.rs
@@ -1,7 +1,10 @@
 use criterion::{black_box, criterion_group, criterion_main, Criterion};
 use lambdaworks_math::{
     cyclic_group::IsGroup,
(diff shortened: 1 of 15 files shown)


MauroToscano:
This is very out of date, we can review an updated version later, I'm closing it

PR #690, opened 2023-11-21
Add Pallas and Vesta fields

mdvillagra:
# Add Pallas and Vesta fields

## Description
This PR add the Pallas and Vesta base fields in order to later construct the Pasta curves. There is no issue associated to this PR, but it was requested in Lambdaworks.

## Type of change

- [x] New feature


MauroToscano (changes requested):
Remove the custom serializer if possible. The one in the Starknet field existed because of a very specific requirement of the the standarized fields of Starknet, and making them work with no std. You don't even need it for wasm.

MauroToscano (approved):

entropidelic (approved):

PR #700, opened 2023-11-27
Feat: Montgomery backend using 32bits

### LINKED ISSUE #538: Implement Montgomery backend with 32 bits
Currently montgomery backend works with limbs of 64 bits. For really small fields, like babybear, it may be useful to have a 32 bits version. This version doesn't need to support multiple limbs, since it's use is limited to this field, but it may be useful to generalize it.

- [ ] Implement a montgomery backend of 32 bits, with only one u32 as limb.
- [ ] Add a fuzzer to compare both montgomery backends
- [ ] Generalize it to be multi limb

We recommend just doing the first task, then generalizing

RajeshRk18:
# Montgomery backend using 32bit words

## Description
This PR resolves #538

## Type of change

- [x] New feature
- [x] Optimization

## Checklist
- [x] Linked to Github Issue
- [x] Unit tests added
  - [ ] Benchmarks added/run


MauroToscano:
This needs to be updated to be merged. It can be re opened if the work continues

PR #716, opened 2023-12-11
Stark: Prover and Verifier over field extensions

schouhy:
# Prover and Verifier over field extensions

## Description

This PR is the first step to support field extensions in the stark prover and verifier. 

In summary, this PR
- Adds associated type `FieldExtension` to both `IsStarkProver` and `IsStarkVerifier`.
- The prover casts the trace to the field extension at the very beginning of the prove method and continues as before.
- Implements polynomial operations between polynomials with coefficients in a field and a subfield. This will be needed in a subsequent PR.

There is a followup PR that improves the way the prover handles the trace and avoids casting it to the field extension. This is so dividided in two PRs to make reviewing easier.

## Type of change

- [x] New feature

### DIFF
--- a/math/src/polynomial.rs
+++ b/math/src/polynomial.rs
@@ -1095,7 +1095,7 @@ mod tests {
         }
     }
 
-    use proptest::prelude::*;
+    
     proptest! {
         #[test]
         fn ruffini_inplace_equals_ruffini(p in any::<Vec<u64>>(), b in any::<u64>()) {
--- a/math/src/polynomial.rs
+++ b/math/src/polynomial.rs
@@ -1095,7 +1095,6 @@ mod tests {
         }
     }
 
-    
     proptest! {
         #[test]
         fn ruffini_inplace_equals_ruffini(p in any::<Vec<u64>>(), b in any::<u64>()) {
--- a/math/src/polynomial.rs
+++ b/math/src/polynomial.rs
@@ -157,7 +157,7 @@ impl<F: IsField> Polynomial<FieldElement<F>> {
                 coefficients.push(c.clone());
                 c = coeff + c * b;
             }
-            coefficients = coefficients.into_iter().rev().collect();
+            coefficients.reverse();
             Polynomial::new(&coefficients)
         } else {
             Polynomial::zero()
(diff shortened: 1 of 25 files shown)


provers/stark/src/transcript.rs
```diff
@@ -15,18 +15,22 @@ pub trait IsStarkTranscript<F: IsField> {
...
+        trace_roots_of_unity: &[FieldElement<S>],
     ) -> FieldElement<F>
     where
         FieldElement<F>: Serializable,
     {
         loop {
             let value: FieldElement<F> = self.sample_field_element();
-            if !lde_roots_of_unity_coset.iter().any(|x| x == &value)
-                && !trace_roots_of_unity.iter().any(|x| x == &value)
+            if !lde_roots_of_unity_coset
```
diegokingston, line 28: Given that roots of unity are real for the cases we have, it would suffice to check if the imaginary part is non-zero to know it is an ood point, avoiding the cloning and extension

diegokingston (approved):

math/src/polynomial.rs
```diff
@@ -158,6 +145,25 @@ impl<F: IsField> Polynomial<FieldElement<F>> {
...
+        F: IsSubFieldOf<L>,
+    {
+        if let Some(c) = self.coefficients.last() {
+            let mut c = c.clone().to_extension();
+            let mut coefficients = Vec::with_capacity(self.degree());
+            for coeff in self.coefficients.iter().rev().skip(1) {
+                coefficients.push(c.clone());
+                c = coeff + c * b;
+            }
+            coefficients = coefficients.into_iter().rev().collect();
```
Oppen, line 160: You can [reverse](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.reverse) the vector in-place:
```suggestion
            coefficients.reverse();
```

provers/stark/src/debug.rs
```diff
@@ -46,7 +51,7 @@ pub fn validate_trace<F: IsFFTField, A: AIR<Field = F>>(
             let boundary_value = constraint.value.clone();
             let trace_value = trace.get(step, col);
 
-            if &boundary_value != trace_value {
+            if &boundary_value.clone().to_extension() != trace_value {
```
MauroToscano, line 54: Is this clone needed ?

MauroToscano (approved):

entropidelic (approved):

PR #717, opened 2023-12-13
Stark: Prover and Verifier over field extensions v2

schouhy:
# Prover and Verifier over field extensions v2

## Description

This is an improvement over PR #716 that avoid casting the trace to the field extension. The immediate consequence is that the prover can no longer consider all trace polynomials and all trace LDE evaluations as a single object since they live in different fields. This PR refactors the prover's and verifier's code to handle those different types.

Evaluation of constraints to build the composition polynomial still considers all columns to live in the field extension. A followup PR will improve that to fully leverage the optimized ops between different fields. This is so divided in two different PRs for making the review process easier.


provers/cairo/tests/wasm.rs
Oppen, line 1: Why does the proof change?

schouhy: It now holds elements in the base field (the openings of the main trace polynomials) and in the extension (evaluations in ood, fri values). I also changed the way we store the openings of the deep poly to be able to hold openings of the main trace polynomials as elements in the base field and elements of the aux trace polynomials as elements of the field extension. That was a refactor that also made some things easier: before we used vectors of lenght 2 to hold the auth paths to the opnenings of the main trace and the aux trace, and now they have their own attributes in the struct.

MauroToscano (approved):
I also checked if there was any performance issue. Proof generation time didn't change. All good

entropidelic (approved):

PR #742, opened 2024-01-04
feat(fiat_shamir): Add ToTranscript trait

PatStiles:
# Adds ToTranscript trait

## Description

Abstracts the logic of appending a witness to the transcript to the struct level.

## Type of change

- [x] New feature

### DIFF
--- a/crypto/src/fiat_shamir/default_transcript.rs
+++ b/crypto/src/fiat_shamir/default_transcript.rs
@@ -1,4 +1,4 @@
-use super::transcript::{Transcript, ToTranscript};
+use super::transcript::{ToTranscript, Transcript};
 use sha3::{Digest, Keccak256};
 
 pub struct DefaultTranscript {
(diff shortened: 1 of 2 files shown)


MauroToscano:
I don't like this API a lot. People need to know there is a to_transcript() trait implemented for most of primitives types. Everyone already knows in Rust to_bytes_le() and to_bytes_be() exists.

For reasons of compatibility we created the `IsStarkTranscript`at some point, that has some specific methods to append and sample structures. Maybe it could be joined with the last one. @entropidelic maybe has another opinion.

MauroToscano:
Also, as I mentioned in the TG, we need multiple ways to append data for compatibility purposes / testing, this locks you with one

crypto/src/fiat_shamir/default_transcript.rs
```diff
@@ -41,7 +41,7 @@ mod tests {
     fn basic_challenge() {
         let mut transcript = DefaultTranscript::new();
 
-        let point_a: Vec<u8> = vec![0xFF, 0xAB];
+        let oint_a: Vec<u8> = vec![0xFF, 0xAB];
```
diegokingston, line 44: Here it should be point_a

PR #748, opened 2024-01-06
fix: typo

Vid201:
# TITLE

## Description

Just a typo fix.


MauroToscano:
Thanks. We are not accepting PRs with typos, to avoid bots, but we will fix this.

PR #751, opened 2024-01-08
Add prover gnark example

MauroToscano:
# Add prover gnark example


Cargo.toml
```diff
@@ -5,7 +5,7 @@ exclude = ["ensure-no_std"]
 resolver = "2"
 
 [workspace.package]
-version = "0.4.0"
+version = "0.5.0"
```
MauroToscano, line 8:
```suggestion
version = "0.4.0"
```

PR #753, opened 2024-01-09
Update Readme + Fix Typo

MauroToscano:
# Update README + Fix Typo

### DIFF
--- a/provers/cairo/README.md
+++ b/provers/cairo/README.md
@@ -8,13 +8,18 @@
 
 This prover is still in development and may contain bugs. It is not intended to be used in production yet.
 
-Please check issues under security label, and wait for them to be resolved if they are relevant to your project.
-
-Output builtin is finished, and range check is supported but it's not sound yet.
-
-CLI currently runs with 100 bits of conjecturable security
+Please check issues under security label, and wait for them to be resolved if they are relevant to your project. Currently builtins are not proven, and checks over public inputs are not enforced.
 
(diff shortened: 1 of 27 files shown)


provers/cairo/README.md
```diff
@@ -8,13 +8,18 @@
 
 This prover is still in development and may contain bugs. It is not intended to be used in production yet.
 
-Please check issues under security label, and wait for them to be resolved if they are relevant to your project.
+Please check issues under security label, and wait for them to be resolved if they are relevant to your project. Currently builtins are not proven, and checks over public inputs are not enforced.
```
entropidelic, line 11: I would say that builtins are not supported rather than not proven. Maybe we can say that we support layout plain, which has no builtins.

Also, I would expand a little more on what "checks over public inputs are not enforced" really means, similar to what appears in the Stone prover README,
> Note: The verifier only checks that the proof is consistent with the public input section that appears in the proof file. The public input section itself is not checked. For example, the verifier does not check what CairoZero program is being proved, or that the builtins memory segments are of valid size. These things need to be checked externally.

MauroToscano: We can expand on the security section, or the readme. I'd not write not supported since the programs can be proven, the builtins won't have the additional constraints.

MauroToscano: I removed everything. Will write the specific issues, else it will bloat the readme

entropidelic (approved):

diegokingston (approved):

PR #757, opened 2024-01-11
Fix typos in trace_plain_layout.md

Thabokani:


MauroToscano:
I appreciate the help, but we are not accepting contributions fixing typos from non contributors

PR #763, opened 2024-01-14
feat: constant time ecc

rahulghangas:
# TITLE
Constant time ecc

## Description
Constant time operations for Elliptic curves, referenced from 
[Constant time addition for elliptic curves](https://eprint.iacr.org/2015/1060)
[Constant time Montgomery Ladder](https://eprint.iacr.org/2020/956) 

Relates to #394 

## Type of change
Implements `ConditionallySelectable` for constant time swaps between all types implementing `IsGroup`
Implement constant time addition for short Weierstrass curve

`Note: This change is currently incompatible with miden types`

- [x] New feature

## Checklist
- [x] Linked to Github Issue


rahulghangas:
I'm not particularly happy with all the `Copy` trait derivations, but there's no way around it unless we move `operate_with_self`'s implementation to concrete types

MauroToscano:
Hello! For future reference, most of the things that are not under the tag "collaborator-friendly" are not things we are expecting to be done, because they usually need more though, so they are more like a remainder of things to check in the future.

This looks good. But we can't drop compatibility with Winterfell. Constant time is not as important as to prove Winterfell for us, since we don't want the library to be used for traditional crypto, and we need to learn more about how to be really sure this is secure since it's not our expertise. 

I'll check this with Diego next week to see what to do

MauroToscano:
In another topic, if you are adding more features, add them to clippy and the CI, so they are tested

rahulghangas:
@MauroToscano ready from my end I think, have added tests to CI and clippy to Makefile as well. Ignoring some trivial things with clippy because they are fine normally but have inconsequential warnings under `constant-time`

MauroToscano:
I have seen some macros. We don't accept macros in the library. We have been avoiding them until now, and we will continue.

rahulghangas:
Got it. Hmm, I can't seem to find a nice way to conditionally define trait bounds without macros, and I don't want to duplicate large blocks of code. Let me see if I can find a clean workaround.

rahulghangas:
@MauroToscano I ended up just duplicating traits, there is no nice way around it unfortunately (unless we try to use specialization, which is still not stable)

rahulghangas:
@MauroToscano bumping this

MauroToscano:
We are having some internals debate on this issue. For the time being let's halt it. Some of the things we are discussing are:

- Finite field operations are not done in constant time, so even if the ECC algorithm is supposed to be constant time,  we are not sure this is going to be enough
- This is increasing the complexity of a module that needs a rework/upgrade to support more coordinates and curve forms
- There's a possibility we can solve this with a better design, instead of just using features, this may be related to the previous commentary. It may be better to upgrade the module and then try this (Maybe we can have some "ConstantTime" point type or something like that)

We need to take think more in how to approach this issue overall. Right now we don't have any suggestion on how to approach issue

PR #811, opened 2024-02-23
feat(ecc): Add Grumpkin curve + refactor fuzzer

PatStiles:
# Add Grumpkin Curve

## Description

This pr adds an implementation of the grumpkin ecc which is a cycle curve over BN254 and used within the Aztec network. Additionally it performs a minor refactor for fuzzer module.

## Type of change

- [x] New feature


diegokingston (approved):

MauroToscano (changes requested):
It needs some test cases generated with Sage or another library

PatStiles:
Does this not qualify? https://github.com/lambdaclass/lambdaworks/pull/811/files#diff-cd24214c4896039a9c9dd13475bcddf6733717c300aa4f42533eb713b25a150dR102 Or do I need to add annotations for each specific test case?

MauroToscano (approved):

PR #812, opened 2024-02-24
Typo in the Monthomery backend `mod` name

skaunov:
Plus some minor editing of docs/comments for readability.


MauroToscano:
We don't accept typo fixes from non contributors. Thanks for reporting though.

Issue #790, opened 2024-01-29
MerkleTree: Add parallelization of inner nodes


MauroToscano:
Merkle Trees are only parallelizing the hashing of the leaves, we should rewrite in a way to allow parallelization of nodes

MauroToscano:
Solved on #824

PR #824, opened 2024-03-01
Merkle trees: Parallelize building of inner levels

irfanbozkurt:
## Description

Addresses #790 by approaching the tree-building process from a level-based approach rather than a recursive one, as it allows treating pairs of children separately to build separate parents for each level, which can totally be parallelized. 

Here's a simple test on M1 chip

```
let leaves: Vec<FE> = (1..u64::pow(2, 30)).map(FE::new).collect();

let mut nodes = vec![FE::zero(); leaves.len() - 1];
nodes.extend(leaves);

let now = Instant::now();
{
    build::<TestBackend<U64PF>>(&mut nodes, ROOT);            // Old interface
    build::<TestBackend<U64PF>>(&mut nodes, leaves.len());    // New interface
}
let elapsed = now.elapsed();
println!("Elapsed: {:.2?}", elapsed);
```
Old implementation: **39.02s**
New implementation without parallel feature: **29.56s**
New implementation with parallel feature: **8.85s**

## Type of change
- [X] Optimization

## Checklist
- [X] Linked to Github Issue
- [X] Unit tests added

### DIFF
--- a/crypto/src/merkle_tree/utils.rs
+++ b/crypto/src/merkle_tree/utils.rs
@@ -1,6 +1,8 @@
 use alloc::vec::Vec;
 
 use super::traits::IsMerkleTreeBackend;
+#[cfg(feature = "parallel")]
+use rayon::prelude::*;
 
 pub fn sibling_index(node_index: usize) -> usize {
     if node_index % 2 == 0 {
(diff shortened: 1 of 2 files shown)


diegokingston (approved):

crypto/src/merkle_tree/utils.rs
```diff
@@ -99,14 +109,14 @@ mod tests {
     const ROOT: usize = 0;
 
     #[test]
-    // expected |10|10|13|3|7|11|2|1|2|3|4|5|6|7|8|
     fn compleate_a_merkle_tree_from_a_set_of_leaves() {
```
MauroToscano, line 112:
```suggestion
fn complete_a_merkle_tree_from_a_set_of_leaves()
```

crypto/src/merkle_tree/utils.rs
```diff
@@ -99,14 +109,14 @@ mod tests {
...
-        let leaves: Vec<FE> = (1..9).map(FE::new).collect();
+        let leaves: Vec<FE> = (0..u64::pow(2, 16)).map(FE::new).collect();
+        let leaf_length = leaves.len();
```
MauroToscano, line 114: leaves_len

MauroToscano (approved):

crypto/src/merkle_tree/utils.rs
```diff
@@ -99,14 +109,14 @@ mod tests {
...
 
-        let mut nodes = vec![FE::zero(); leaves.len() - 1];
+        let mut nodes = vec![FE::zero(); leaves_len - 1];
         nodes.extend(leaves);
 
-        build::<TestBackend<U64PF>>(&mut nodes, ROOT);
-        assert_eq!(nodes[ROOT], FE::new(10));
+        build::<TestBackend<U64PF>>(&mut nodes, leaves_len);
+        assert_eq!(nodes[ROOT], FE::new(3));
     }
```
MauroToscano, line 121: Re add this test case, I can't check the numbers quickly on the new one.

MauroToscano (changes requested):
The new test result can't be checked. Restore the old test case. If the new one is needed for some reason keep it too, try to add the results of the mod 13 of the pow

MauroToscano (approved):

PR #828, opened 2024-03-01
Use `ShrAssign` for exponent in `pow` instead of `Shr`

tcoratger:
# Use `ShrAssign` for exponent in `pow` instead of `Shr`

## Description

Currently, the `pow` function utilizes the `Shr` method to right-shift the exponent, which creates a new value each time it's invoked. This approach may be slightly less efficient than using `ShrAssign`, which directly modifies the original exponent in place. By replacing `Shr` with `ShrAssign`, we can potentially improve the performance of the `pow` function, especially for larger exponents.

## Type of change

- [X] Optimization

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run


MauroToscano:
Can you fix clippy ?

diegokingston (approved):

MauroToscano (approved):

Issue #830, opened 2024-03-06
Add Jacobian Coordinates for Short Weierstrass


MauroToscano:
Short Weierstrass is the most used EC form of our library. Lambdaworks should support jacobian coordinates that are faster

- [ ] Add JacobianPoint in short weierstrass point.rs .Implement the `IsGroup`trait, partial_eq, and add a to_affine method. Use the same API for the new function. 
- [ ] Add serialization functions
- [ ] Add conversion with from to move from both point formats

A draft of the algorithms can be found here:

https://github.com/lambdaclass/lambdaworks/pull/619/files

skaunov:
I'm starting it! X)

MauroToscano:
Perfect

PR #838, opened 2024-03-07
#830 - Jacobian Points

skaunov:
# TITLE

## Description

See #830

## Type of change

- [ ] New feature

## Checklist
- [x] Linked to Github Issue
- [x] Unit tests added
  - [ ] Benchmarks added/run

### DIFF
--- a/math/src/elliptic_curve/short_weierstrass/point/jacobian.rs
+++ b/math/src/elliptic_curve/short_weierstrass/point/jacobian.rs
@@ -1,5 +1,7 @@
 use crate::{cyclic_group::IsGroupElement, elliptic_curve::{short_weierstrass::traits::IsShortWeierstrass, traits::IsEllipticCurve}, field::element::FieldElement};
 
+use super::ShortWeierstrassProjectivePoint;
+
 // feels like `crate::elliptic_curbe::point` could be redesigned to provide the trait to implement here
 #[derive(Clone)]
 pub struct JacobianPoint<E: IsShortWeierstrass> {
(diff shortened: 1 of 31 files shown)


MauroToscano:
There are some conflicts here

skaunov:
It's more surprising not to have them while active developing.

PS 
this one in particular due to the renaming you closed earlier I guess
but I won't be surprised by new conflicts pop up before positive review

math/src/elliptic_curve/short_weierstrass/point/jacobian.rs
```diff
@@ -0,0 +1,132 @@
...
+        }
+    }
+    /// Creates an elliptic curve point giving the Jacobian projective $[x: y: 1]$ coordinates.
+    pub const fn new(
+        x: FieldElement<E::BaseField>,
+        y: FieldElement<E::BaseField>,
+        z: FieldElement<E::BaseField>,
+    ) -> Self {
+        Self { x, y, z }
+    }
```
MauroToscano, line 62: We should make the other new in projectives work like this too, I think it's easier to use. It can be made in a followup

MauroToscano (changes requested):
Tests need to be added. It's looking good

skaunov:
I was planning to adapt tests from projective, but found no actual, only for bytes conversions IIRC.

* What's an other fine source of a test vectors suite?
* Is it ok if projective lacks actual tests?

MauroToscano:
> I was planning to adapt tests from projective, but found no actual, only for bytes conversions IIRC.
> 
>     * What's an other fine source of a test vectors suite?
> 
>     * Is it ok if projective lacks actual tests?

We don't have the tests in the general coordinates, because by their own they don't anything without a curve. But you could add it to the tests of the bls 12 381 for example, and test jacobians for that specific curve.

skaunov:
I'm planning to use affine for `tests` which will be testing if the results here match `Projective` after both reduced to affine; plus cover the other added functions.

Thing is I have couple of things which will end in couple of weeks, after which I would be happy to finish this one. If anything is really demanding/depending on this I can prioritize this.

I has the feeling that this one became outstanding and urges for an action just in spite of tidiness. If it's the case maybe we have an appropriate label here, or should we convert it to be draft for a couple of weeks?

skaunov:
I feel it was a mistake from to re-request review from Mauro... Sorry. =(

MauroToscano:
We are not in a hurry to finish it. We can wait. If for some reason you will abandon it, then do tell please.

skaunov:
Of course I would. But this one is finished with the last commit. I 
notified Irfan via Tg that it's ready for the final review and 
squashing&merging.

diegokingston:
This functionality has already been added

PR #840, opened 2024-03-09
Adjust imports to target Rust nightly

tcoratger:
# Adjust imports to target Rust nightly

## Description

New components such as `Vec` or `ToString` have recently been added to the standard library prelude (`std`). As Lambdaworks is predominantly `no_std`, importing `::vec::Vec` explicitly leads to Rust interpreting it as an import overlay, considering that `Vec` already exists in the `std` prelude. However, without the `no_std` flag, Rust fails to recognize `Vec` due to its inability to access the `std` prelude. Consequently, to avoid warnings and ensure compatibility, imports have been modified to use `vec::*`.

## Type of change

- [X] Optimization

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run

### DIFF
--- a/crypto/src/commitments/kzg.rs
+++ b/crypto/src/commitments/kzg.rs
@@ -1,5 +1,5 @@
 use super::traits::IsCommitmentScheme;
-use alloc::{borrow::ToOwned, vec::Vec};
+use alloc::{borrow::*, vec::*};
 use core::{marker::PhantomData, mem};
 use lambdaworks_math::{
     cyclic_group::IsGroup,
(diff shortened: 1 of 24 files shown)


MauroToscano:
Can you show which warning are you fixing ? I'm not getting why is this an improvement, and it will come back to the library because people will just import what they need if they see no warning/error

MauroToscano:
We will upgrade whenever the nightly changes come to the stable version of rust, if we merge this now it will break in the future since we don't have checks in the CI

PR #842, opened 2024-03-11
Small perf adjustments for `ShortWeierstrassProjectivePoint` operations

tcoratger:
# Small perf adjustments for `ShortWeierstrassProjectivePoint` operations

## Description

- In the `double` operation, the pull request implements an early return strategy to check if `self` is the neutral element before performing any further calculations. If `self` is indeed the neutral element, the function immediately returns a clone of `self`. This optimization enhances performance by avoiding unnecessary computations, resulting in faster execution.

- The pull request optimizes the `operate_with_affine` function by reordering the sequence of operations. Previously, the function performed certain calculations before checking if `self` or `other` were neutral elements. This resulted in unnecessary computation overhead. The pull request rectifies this by placing the neutral element check at the beginning of the function. If either `self` or `other` is the neutral element, the function terminates early, saving computational costs and improving overall efficiency.

## Type of change

- [X] Optimization

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run


MauroToscano:
I don't think the first optimization is right. The early return adds an additional branch, so it may make the average case slower, and we usually don't care about optimizing the neutral element case. You need to bench this and compare it, but I highly doubt the additional branch is worth.

tcoratger:
> I don't think the first optimization is right. The early return adds an additional branch, so it may make the average case slower, and we usually don't care about optimizing the neutral element case. You need to bench this and compare it, but I highly doubt the additional branch is worth.

Here is the benchmark for the double case (for the general situation):
- Before:
```shell
point Projective Double | Lambdaworks
                        time:   [364.85 ns 364.97 ns 365.09 ns]
Found 7 outliers among 100 measurements (7.00%)
  5 (5.00%) high mild
  2 (2.00%) high severe
```

- After:
```shell
point Projective Double | Lambdaworks
                        time:   [364.55 ns 365.17 ns 366.33 ns]
                        change: [-0.1313% +0.0010% +0.1993%] (p = 1.00 > 0.05)
                        No change in performance detected.
Found 10 outliers among 100 measurements (10.00%)
  4 (4.00%) high mild
  6 (6.00%) high severe
```

MauroToscano (approved):

entropidelic (approved):

tcoratger:
@entropidelic Let me know if there is something to do here

Oppen (approved):

PR #855, opened 2024-04-03
Poseidon optimization starknet

MauroToscano:
# Poseidon optimization

## Description

Enables the mix function to be redefined in the PoseidonParams. This enables optimization when there's a better alternatively to doing the naive permutation, when the MDS parameters were picked with that purpose in mind. 

In that case, the MDS parameters maybe dummies, but we keep the static initialization to make Poseidon easily to instantiate by default.

I have also allowed op ref, because field elements should be used with a reference, as to avoid copying them, for efficiency. This gives a 2% speedup in this case. This clippy suggestion leads to worse efficiency

## Type of change

- [x] New feature
- [x] Optimization

# Benchmarks

These are the results of the benchmarks against our previous naive implementation:

image


crypto/src/hash/poseidon/starknet/parameters.rs
```diff
@@ -41,7 +51,7 @@ impl PoseidonCairoStark252 {
     // And the round 0 ones matches the one used
     // in Cairo Lang
     // https://github.com/starkware-libs/cairo-lang/blob/c98fc0b50529185b7018208cb3460191eeb53e0d/src/starkware/cairo/stark_verifier/air/layouts/starknet/autogenerated.cairo#L1574-L1596
-    const ROUND_CONSTANTS: [FE<Stark252PrimeField>; 3 * 91] = [
+    pub const ROUND_CONSTANTS: [FE<Stark252PrimeField>; 3 * 91] = [
```
Oppen, line 54: If it's not used I'd rather rename it to start with `_`. Wdyt?

MauroToscano: Round Constants are used. For the case of Starknet we don't use the MDS matrix and we apply a simplification of it

MauroToscano: Leaving aside that, I don't think we can do it, because for the generic case we use them, only in the optimized case we don't. Maybe there's a way to embed something with another trait the solves the mix either with constants or with matrixes, but I fear adding new implementations with other constants will be cumbersome that way

entropidelic (approved):

diegokingston (approved):

PR #866, opened 2024-04-21
Add `to_bits_be` for `FieldElement<Stark252PrimeField>`

tcoratger:
# Add `to_bits_be` for `FieldElement<Stark252PrimeField>`

## Description

Following `to_bits_le`, an implementation of `to_bits_be` for `FieldElement<Stark252PrimeField>` is added. A unit test is also integrated to validate the correct implementation of the two methods.

## Type of change

- [X] New feature
- [ ] Bug fix
- [ ] Optimization

## Checklist
- [ ] Linked to Github Issue
- [X] Unit tests added
- [X] This change requires new documentation.
  - [X] Documentation has been added/updated.
  - [ ] Benchmarks added/run


MauroToscano:
Bits operations in field elements is a bad idea. In the base type it's ok, if you want to use it you should use the raw data of the base type or call the representative. Calling the representative all the time means a heavy penalty to the operation cost, so it's not a good idea to use it

MauroToscano:
I see now that we have the to bits_le, we should probably remove both in the future, in the meanwhile we can add it

diegokingston:
We will be removing the functionality of to_bits_le, so I am closing this PR

PR #876, opened 2024-06-21
feat(merkle): Pad single tree leaves to next power of two

PatStiles:
# Pad Single Tree leaves to next power of two

## Description

For merkle trees with a single leaf, pads the number of leaves to the next power of two by appending the last element.

## Type of change

- [x] New feature

## Checklist
- [x] Unit tests added

### DIFF
--- a/crypto/src/merkle_tree/utils.rs
+++ b/crypto/src/merkle_tree/utils.rs
@@ -28,14 +28,12 @@ pub fn complete_until_power_of_two<T: Clone>(values: &mut Vec<T>) -> Vec<T> {
     values.to_vec()
 }
 
-/*
-    ! NOTE !
-    - We in this function we  say 2^0 = 1 is not a power of two when it is infact true. We 
-    need this to maintain that the smallest tree at two leaves counts and we could not find 
-    a name that wasn't overly verbose. In the case of a single value being present in a leaf node 
-    this indicates we pad to next power of two (i.e. 2). The function is private and is only used 
-    to ensure the tree has a power of 2 number of leaves.
-*/
+// ! NOTE !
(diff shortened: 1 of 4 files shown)


MauroToscano (approved):

crypto/src/merkle_tree/utils.rs
```diff
@@ -22,17 +22,18 @@ pub fn parent_index(node_index: usize) -> usize {
...
     values.to_vec()
 }
 
 pub fn is_power_of_two(x: usize) -> bool {
-    (x != 0) && ((x & (x - 1)) == 0)
+    if x == 1 {
+        false
+    } else {
+        (x != 0) && ((x & (x - 1)) == 0)
+    }
```
Oppen, line 36: Strictly speaking, this is not true. `1` is `2^0`, so it is indeed a power of `2`. That said, here's a simpler alternative:
```suggestion
    (x > 1) && ((x & (x - 1)) == 0)
```

Oppen: Because technically we're lying, I expect a comment clarifying that:
1. We need this to keep the smallest tree at two leaves count.
2. We couldn't find a better name that wasn't overly verbose to what the function really does.
3. The function is private (please make it private as well) and is only used to ensure the tree has a power of 2 number of leaves.
4. A little white lie here saves us a lot of edge cases all around the code.

Otherwise, in a few months we will look at this, see it lies, conclude it's a bug, and proceed to "fix" it causing unexplainable bugs elsewhere.

PatStiles: Thanks @Oppen will address this now.

PR #879, opened 2024-06-26
chore(dep): Pin starknet-rs dependencies

PatStiles:
# Pins starknet-rs dependencies

## Description

Pins starknet-rs dependencies.

## Type of change
- [x] Bug fix


Oppen (approved):

Oppen:
@diegokingston @MauroToscano we'll need to merge this with IAI not running. The cause is main is broken due to unpinned deps in the benchmarks (this PR fixes it). The first step to IAI benchmarks is running for main, which fails for the aforementioned issue. The `CI / Compile` workflow checks that the current version of benchmarks build, so once in main other PRs should start passing the IAI step.

diegokingston (approved):

PR #884, opened 2024-07-24
Add lambdaworks Pinocchio example

nicole-graus:
# Pinocchio example

## Description

Add lambdaworks Pinocchio example based on the previous implementation.


examples/pinocchio/Readme.md
```diff
@@ -0,0 +1,37 @@
+# [WIP]
+# Lambdaworks Pinocchio 
+
+This is an implementation of [Pinocchio protocol](https://eprint.iacr.org/2013/279.pdf) using Lambdaworks. This sourse code is the companion of this [blog post](link) aimed at those who want to learn about SNARKs.
+
+# Understanding the code
+We encourage to start by reading the [blog post](link) to understand the code.
```
Oppen, line 7: I suggest updating the link with the one the post will have when published.

examples/pinocchio/Readme.md
```diff
@@ -0,0 +1,37 @@
...
+
+Then, in the `tests/` module you will find integration tests that will guide you through the happy path: the setup, the prover and the verifier. Each of these components can be located in their own files:
+
+- `pinocchio/setup.rs`: generates the `VerificationKey` and the `EvaluationKey` with the relevant data to construct and verify proofs. This data comes from the structure of the program P encoded as `Polynomials` in a `QAP` (Quadratic arithmetic program). To hide this data, random `FieldElement`s are sampled as `ToxicWaste` and then mapped to `G1Point`'s and  `G1Point`'s via repeated addition of the `generator()`'s of the curves.
```
Oppen, line 11: I think there's a typo:
```suggestion
- `pinocchio/setup.rs`: generates the `VerificationKey` and the `EvaluationKey` with the relevant data to construct and verify proofs. This data comes from the structure of the program P encoded as `Polynomials` in a `QAP` (Quadratic arithmetic program). To hide this data, random `FieldElement`s are sampled as `ToxicWaste` and then mapped to `G1Point`'s and  `G2Point`'s via repeated addition of the `generator()`'s of the curves. 
```

examples/pinocchio/src/common.rs
```diff
@@ -0,0 +1,36 @@
+use lambdaworks_math::{
+    elliptic_curve::{
+        short_weierstrass::curves::bls12_381::{
+            curve::BLS12381Curve, default_types::FrElement, default_types::FrField as FrF,
```
Oppen, line 4: Given you're aliasing `FrField` later, you may just import with the real name:
```suggestion
            curve::BLS12381Curve, default_types::FrElement, default_types::FrField,
```

examples/pinocchio/src/common.rs
```diff
@@ -0,0 +1,36 @@
...
+    },
+    unsigned_integer::element::U256,
+};
+use rand::{Rng, SeedableRng};
+
+pub type Curve = BLS12381Curve;
+pub type TwistedCurve = BLS12381TwistCurve;
+
+pub type FE = FrElement;
+pub type F = FrF;
```
Oppen, line 17: To match the above:
```suggestion
pub type F = FrField;
```

examples/pinocchio/src/common.rs
```diff
@@ -0,0 +1,36 @@
...
+pub type Pairing = BLS12381AtePairing;
+
+pub type G1Point = <BLS12381Curve as IsEllipticCurve>::PointRepresentation;
+pub type G2Point = <BLS12381TwistCurve as IsEllipticCurve>::PointRepresentation;
+
+pub const ORDER_R_MINUS_1_ROOT_UNITY: FrElement = FrElement::from_hex_unchecked("7");
+
+pub fn sample_fr_elem() -> FrElement {
+    let mut rng = rand_chacha::ChaCha20Rng::seed_from_u64(9001);
+    FrElement::new(U256 {
```
Oppen, line 28: For consistency, I suggest using the type alias here:
```suggestion
pub const ORDER_R_MINUS_1_ROOT_UNITY: FE = FE::from_hex_unchecked("7");

pub fn sample_fr_elem() -> FE {
    let mut rng = rand_chacha::ChaCha20Rng::seed_from_u64(9001);
    FE::new(U256 {
```

examples/pinocchio/src/qap.rs
```diff
@@ -0,0 +1,362 @@
...
+    /// outputs associated v,w,y polynomials
+    pub fn new(
+        vs: Vec<Polynomial<FE>>,
+        ws: Vec<Polynomial<FE>>,
+        ys: Vec<Polynomial<FE>>,
+        target: Polynomial<FE>,
+        number_of_inputs: usize,
+        number_of_outputs: usize,
+    ) -> Result<Self, CreationError> {
+        if vs.len() != ws.len() || vs.len() != ys.len() || ws.len() != ys.len() {
```
Oppen, line 36: Nit: the third clause is unreachable here, since `ws.len() != ys.len()` implies at least one of them needs to also be `!= vs.len()` first.

examples/pinocchio/src/qap.rs
```diff
@@ -0,0 +1,362 @@
...
+            Err(CreationError::PolynomialVectorsSizeMismatch)
+        } else {
+            Ok(Self {
+                vs,
+                ws,
+                ys,
+                target,
+                number_of_inputs,
+                number_of_outputs,
```
Oppen, line 45: If I understand correctly, there's also a condition that `number_of_inputs + number_of_outputs <= vs.len()`, is that correct?

examples/pinocchio/Readme.md
diegokingston: source code

Oppen (approved):

diegokingston (approved):

juanbono (approved):

PR #913, opened 2024-09-14
Add explicit drop of `toxic_waste` once SRS is generated

0xAdriaTorralba:
# Optimization changes to `kzg.rs`

## Description

The aim of this PR is to improve the `kzg.rs` file.

In particular, I have added two optimizations:

1. Added `MAX_POLYNOMIAL_DEGREE` as a constant to determine the maximum degree of polynomial that is allowed to commit.
2. Added explicit `drop` of memory of the randomly generated `toxic_waste` once the SRS is generated. Although Rust manages the memory (and in particular, the drops of unused variables) I find particularly important to handle the deletion of this variable *explicitly* and *manually*. Since, it is CRUCIAL for the soundness of the protocol for this variable to be DELETED right after its job is done.

## Type of change

- [x] Optimization

## Checklist
- [ ] ~~Linked to Github Issue~~
- [x] Unit tests added
- [ ] ~~This change requires new documentation.~~
  - [ ] ~~Documentation has been added/updated.~~
  - [ ] ~~Benchmarks added/run~~

### DIFF
--- a/crypto/src/commitments/kzg.rs
+++ b/crypto/src/commitments/kzg.rs
@@ -314,6 +314,82 @@ mod tests {
     }
 
     #[test]
+    fn e2e_kzg_single_polynomial() {
+        // Setup phase: create the SRS
+        let kzg = KZG::new(create_srs());
+
+        // Commit phase: create the desired polynomial, and commit that polyonmial
+        let p = Polynomial::<FrElement>::new(&[
+            FieldElement::from(7),
+            FieldElement::from(5),
+            FieldElement::from(3),
+        ]);
+        // ASSERT: the polynomial is not too large
+        assert!(p.degree() <= MAX_POLYNOMIAL_DEGREE);
+        let p_commitment: <BLS12381AtePairing as IsPairing>::G1Point = kzg.commit(&p);
+
+        // Open phase: evaluate the polynomial at the desired point i, and create the proof for that opening
+        let i = FieldElement::from(1); // Challenge from the verifier
+        let y = p.evaluate(&i); // Evaluation of the polynomial at challenge
+        // ASSERT: the evaluation is correct
+        assert_eq!(y, FieldElement::from(15));
+        let proof = kzg.open(&i, &y, &p);
+
+        // Verify phase: the proof should verify
+        // ASSERT: the proof verifies
+        assert!(kzg.verify(&i, &y, &p_commitment, &proof));
+    }
+
+    #[test]
+    fn e2e_kzg_batched_polynomials() {
+        // Setup phase: create the SRS
+        let kzg = KZG::new(create_srs());
+        
+        // Commit phase: create the polynomials, and compute the commitments.
+        let p0 = Polynomial::<FrElement>::new(&[
+            FieldElement::from(5),
+            FieldElement::from(4),
+            -FieldElement::from(7),
+        ]);
+        // ASSERT: deg(p0) < MAX_POLYNOMIAL_DEGREE
+        assert!(p0.degree() < MAX_POLYNOMIAL_DEGREE);
+        let p0_commitment: <BLS12381AtePairing as IsPairing>::G1Point = kzg.commit(&p0);
+        
+        let p1 = Polynomial::<FrElement>::new(&[
+            FieldElement::from(1),
+            FieldElement::from(2),
+            -FieldElement::from(1),
+        ]);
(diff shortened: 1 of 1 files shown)


Oppen:
IIUC, the goal of this PR is to remove secrets from memory as soon as they stop being needed, right? Can you confirm this?

If it is, I'm afraid `mem::drop` doesn't do what we need, since `drop` doesn't clean the values, it just releases the resources and makes the address inaccessible in _safe_ Rust, which an attacker would not restrict to. You would need an implementation of `Drop` that `memset`s to `0` (or some poison value) so the value does not linger. There are some crates that do that for you IIRC.
But regardless of that, it seems the changes are just to a test routine that only gets built for tests, it sounds unlikely that there is any meaningful secrets there.
The code certainly does not seem wrong and the extra tests are appreciated, but it doesn't seem as critical as the description implies.
Lastly, the change is labeled as an optimization, but it doesn't seem like one (other than slightly decreasing the lifetime of one value on the stack) and no benchmarks are provided. I would remove that claim unless there is evidence that it improves performance of some important operation.

0xAdriaTorralba:
> IIUC, the goal of this PR is to remove secrets from memory as soon as they stop being needed, right? Can you confirm this?

Yes, that's the goal of this PR.

> If it is, I'm afraid mem::drop doesn't do what we need, since drop doesn't clean the values, it just releases the resources and makes the address inaccessible in safe Rust, which an attacker would not restrict to. You would need an implementation of Drop that memsets to 0 (or some poison value) so the value does not linger. There are some crates that do that for you IIRC.

I didn't knew about this. I'll modify the PR to adapt to this needs.
 
> But regardless of that, it seems the changes are just to a test routine that only gets built for tests, it sounds unlikely that there is any meaningful secrets there.

I know that this routine is intended only for test (and learning) pruposes. However, any other person (just like me) could be interested on learning more about this protocol, and I find that deleting the toxic waste *explicitly* after its usage will help and reinforce the importance of this asset being properly handled and disposed.

> The code certainly does not seem wrong and the extra tests are appreciated, but it doesn't seem as critical as the description implies.

I apologize for the alarm. I understand that this is a code inteded for testing and learning purposes.

> Lastly, the change is labeled as an optimization, but it doesn't seem like one (other than slightly decreasing the lifetime of one value on the stack) and no benchmarks are provided. I would remove that claim unless there is evidence that it improves performance of some important operation.

Okay, what label do you find more suitable for this PR?

Oppen:
WTF did I touch? I meant to quote you, not edit your comment. Sorry!

0xAdriaTorralba:
> WTF did I touch? I meant to quote you, not edit your comment. Sorry!

Don't worry, I'm going to try to recreate the original message.

---

> IIUC, the goal of this PR is to remove secrets from memory as soon as they stop being needed, right? Can you confirm this?

Yes, this is the purpose of this PR.

> If it is, I'm afraid mem::drop doesn't do what we need, since drop doesn't clean the values, it just releases the resources and makes the address inaccessible in safe Rust, which an attacker would not restrict to. You would need an implementation of Drop that memsets to 0 (or some poison value) so the value does not linger. There are some crates that do that for you IIRC.

I didn't know about this. I'll take the opportunity to submit a modification taking this into account.

> But regardless of that, it seems the changes are just to a test routine that only gets built for tests, it sounds unlikely that there are any meaningful secrets there.

I know that this only affects the test routine and that it should not store any meaningful secret. However, I found this code very self-explanatory and very suitable to have a hands-on experience with KZG commitments. For this reason, I find it important to *explicitly* and *manually* remove the toxic waste once it's used, to exemplify good behaviour when implementing a production-ready KZG commitment.

> The code certainly does not seem wrong and the extra tests are appreciated, but it doesn't seem as critical as the description implies.

Regarding the extra tests, that's great. On the other hand, I know that this is not a critical affair right now, I'm sorry if it sounded too alarmist.

> Lastly, the change is labeled as an optimization, but it doesn't seem like one (other than slightly decreasing the lifetime of one value on the stack) and no benchmarks are provided. I would remove that claim unless there is evidence that it improves the performance of some important operation.

I apologize for that, I did not find any other label in which this would be better fitted. What do you propose?

diegokingston:
This PR is stale, I am closing it

0xAdriaTorralba:
Maybe you can reconsider it and check the updated version with the feedback provided... or at least say why this is not suitable for merge

PR #919, opened 2024-09-26
feat: Allow merkle trees of 1 node (its root)

uri-99:
# Allow merkle trees of 1 node

## Description

This PR removes the constraint from `is_power_of_two()` that said 1 is not a power of 2.

In turn, this allows Merkle Trees of size 1 to be built, without padding them to 2 nodes. Size 1 Merkle Trees are then only 1 node, which is equal to its root. 

## Type of change

- [x] New feature: Allow Merkle Trees of size 1

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
- [x] Unit tests updated
  - [ ] Benchmarks added/run


diegokingston (approved):

crypto/src/merkle_tree/proof.rs
```diff
@@ -163,8 +163,8 @@ mod tests {
         let merkle_tree = MerkleTree::<TestBackend<U64PF>>::build(&values).unwrap();
 
         // Update the expected root value based on the actual logic of TestBackend
-        // For example, if combining two `1`s results in `4`, update this accordingly
-        let expected_root = FE::new(4); // Assuming combining two `1`s results in `4`
+        // For example, in this case hashing a singe `1` results in `2`
```
Oppen, line 166:
```suggestion
        // For example, in this case hashing a single `1` results in `2`
```

pablodeymo (approved):
LGTM

PR #930, opened 2024-10-16
Add Rescue Prime Optimized

jotabulacios:
# Add Rescue Prime Optimized

## Description

This PR adds the Rescue Prime Optimized  hash function.
The implementation  offers  three types of MDS transformations: Matrix Multiplication, Karatsuba, and NTT

Example:
 
```rust
let rescue = RescuePrimeOptimized::new(SecurityLevel::Sec128, MdsMethod::Ntt).unwrap();
let input = b"Hello there";
let hash_output = rescue.hash(input);
```

In this case `128` is the security level, the other available level is `160`
To use other `MdsMethod` simply change `MatrixMultiplication` for  `Karatsuba` or `Ntt`
## Type of change

- [ ] New feature


crypto/src/hash/rescue_prime/mod.rs
```diff
@@ -0,0 +1,816 @@
...
+type Fp = FieldElement<Goldilocks64Field>;
+
+#[derive(Clone)]
+#[allow(dead_code)]
+enum MdsMethod {
+    MatrixMultiplication,
+    Ntt,
+    Karatsuba,
+}
+#[warn(dead_code)]
```
pablodeymo, line 21: why do you need this annotation?

crypto/src/hash/rescue_prime/mod.rs
```diff
@@ -0,0 +1,816 @@
...
+
+    pub fn apply_inverse_sbox(state: &mut [Fp]) {
+        for x in state.iter_mut() {
+            *x = x.pow(ALPHA_INV);
+        }
+    }
+
+    fn get_mds_vector(m: usize) -> Vec<Fp> {
+        match m {
+            12 => vec![7, 23, 8, 26, 13, 10, 9, 7, 6, 22, 21, 8]
```
pablodeymo, line 80: this can be an array instead of a vec

crypto/src/hash/rescue_prime/mod.rs
```diff
@@ -0,0 +1,816 @@
...
+                .map(Fp::from)
+                .collect(),
+            16 => vec![
+                256, 2, 1073741824, 2048, 16777216, 128, 8, 16, 524288, 4194304, 1, 268435456, 1,
+                1024, 2, 8192,
+            ]
+            .into_iter()
+            .map(Fp::from)
+            .collect(),
+            _ => panic!("Unsupported state size"),
```
pablodeymo, line 91: I think this function shouldn't panic. It can return a Result indicating the error case.

crypto/src/hash/rescue_prime/mod.rs
```diff
@@ -0,0 +1,816 @@
...
+            MdsMethod::Karatsuba => self.mds_karatsuba(state),
+        };
+        state.copy_from_slice(&new_state);
+    }
+
+    fn add_round_constants(&self, state: &mut [Fp], round: usize) {
+        let m = self.m;
+        let round_constants = &self.round_constants;
+        for j in 0..m {
+            state[j] += round_constants[round * 2 * m + j];
```
pablodeymo, line 185: instead of indexing, this can be improved by using get_mut and an iterator

crypto/src/hash/rescue_prime/utils.rs
```diff
@@ -0,0 +1,75 @@
...
+            let mut buf = [0u8; 8];
+            buf[..chunk.len()].copy_from_slice(chunk);
+            if chunk.len() < 7 {
+                buf[chunk.len()] = 1;
+            }
+            let value = u64::from_le_bytes(buf);
+            Fp::from(value)
+        })
+        .collect()
+}
```
pablodeymo, line 17: Is this function just for rescue_prime or a function that can be used elsewhere?

crypto/src/hash/rescue_prime/mod.rs
```diff
@@ -0,0 +1,816 @@
+use crate::alloc::vec::Vec;
```
pablodeymo, line 1: It's better not to have a giant mod.rs file.
Instead, it's better to have a separate file with each struct or enum.

crypto/src/hash/rescue_prime/mod.rs
```diff
@@ -0,0 +1,816 @@
...
+
+#[derive(Clone)]
+#[allow(dead_code)]
+enum MdsMethod {
+    MatrixMultiplication,
+    Ntt,
+    Karatsuba,
+}
+#[warn(dead_code)]
+pub struct RescuePrimeOptimized<const SECURITY_LEVEL: usize, const NUM_FULL_ROUNDS: usize> {
```
ilitteri, line 22: Is it strictly necessary for the `SECURITY_LEVEL` to be a const generic? Can't we have a `security_level: SecurityLevel` field in `RescuePrimeOptimized`? This would simplify some logic and help remove the `assert` in `RescuePrimeOptimized::new`. If we cannot, other ways exist to simplify said logic.

From what I could see from this code, this is not enough justified (but I could be missing some context).

crypto/src/hash/rescue_prime/mod.rs
```diff
@@ -0,0 +1,816 @@
...
+    fn default() -> Self {
+        Self::new(MdsMethod::MatrixMultiplication)
+    }
+}
+
+impl<const SECURITY_LEVEL: usize, const NUM_FULL_ROUNDS: usize>
+    RescuePrimeOptimized<SECURITY_LEVEL, NUM_FULL_ROUNDS>
+{
+    fn new(mds_method: MdsMethod) -> Self {
+        assert!(SECURITY_LEVEL == 128 || SECURITY_LEVEL == 160);
```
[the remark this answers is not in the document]
ilitteri: I just found the `SecurityLevel` enum. Let's better compare the `SECURITY_LEVEL` with the enum instead.

crypto/src/hash/rescue_prime/mod.rs
ilitteri, line 80: In addition to the comment above, seeing that these values are constant I'd add them as `Fp` to avoid the `.into_iter().map(Fp::from).collect()`.

Ditto for `m = 16`.

crypto/src/hash/rescue_prime/mod.rs
```diff
@@ -0,0 +1,816 @@
...
+            ]
+            .into_iter()
+            .map(Fp::from)
+            .collect(),
+            _ => panic!("Unsupported state size"),
+        }
+    }
+
+    fn generate_circulant_matrix(mds_vector: &[Fp]) -> Vec<Vec<Fp>> {
+        let m = mds_vector.len();
```
ilitteri, line 96: Does this `m` have to do something with the `m` that represents a length that can only be `12` or `16`? If so, don't we need to check its size? And if not, I'd use a more descriptive name for it.

crypto/src/hash/rescue_prime/mod.rs
```diff
@@ -0,0 +1,816 @@
...
+            self.apply_mds(state);
+            self.add_round_constants_second(state, round);
+            Self::apply_inverse_sbox(state);
+        }
+    }
+
+    pub fn hash(&self, input_sequence: &[Fp]) -> Vec<Fp> {
+        let m = self.m;
+        let capacity = self.capacity;
+        let rate = self.rate;
```
ilitteri, line 212: We don't need these variable declarations here (we do not have any advantage in doing this).

crypto/src/hash/rescue_prime/parameters.rs
```diff
@@ -0,0 +1,416 @@
...
+    Fp::const_from_raw(10869617590371203777),
+    Fp::const_from_raw(4582902440098948914),
+];
+
+pub enum SecurityLevel {
+    Sec128,
+    Sec160,
+}
+
+pub fn get_round_constants(level: SecurityLevel) -> &'static [Fp] {
```
ilitteri, line 411: Why return `&'static [Fp]` instead of `[Fp]`?

crypto/src/hash/rescue_prime/utils.rs
ilitteri, line 1: We should reference any documentation/implementation/paper if any of these algorithms were implemented based on it.

crypto/src/hash/rescue_prime/mod.rs
ilitteri: +1

ilitteri (commented):
As a general rule, we won't want to panic anywhere. As this is a library meant for users, we must provide them `Result`s of the operations to let them decide what to do with the errors in such cases. Commonly, panics can happen in places where unwraps, expects, as conversions, and slice indexing (and obviously when we explicitly use the panic macro). Unless it is planned for another PR to handle these cases I'd suggest doing it asap.

Unwraps and expects can be avoided returning Results. As conversions can be avoided mostly by using `.into()`, and `.try_into()` (we will prefer the first one as we'd not have to handle the error). Finally, slice indexing can be avoided using `.get()` and handling the Option. 

To improve the code's readability we'd want to define descriptive constants for hardcoded structures.

crypto/src/hash/rescue_prime/parameters.rs
jotabulacios: Using `[Fp] `leads to compile error since we return arrays of two different sizes.  ` &'static [Fp]` seems to solve it

crypto/src/hash/rescue_prime/mod.rs
jotabulacios: Solved, it wasn’t necessary. I added it while trying to fix some Clippy issues, but it turned out to be incorrect

crypto/src/hash/rescue_prime/mod.rs
jotabulacios: This function isn't used anymore, but I acknowledged your comment about avoiding panics

crypto/src/hash/rescue_prime/mod.rs
jotabulacios: updated the implementation as suggested

crypto/src/hash/rescue_prime/utils.rs
jotabulacios: Currently it 's only use in rescue_prime. Maybe later it can be added to `fields`

crypto/src/hash/rescue_prime/mod.rs
jotabulacios: Restructured the files as suggested. Now is more maintainable and clear

crypto/src/hash/rescue_prime/mod.rs
jotabulacios: This vector now is precomputed in `parameters.rs` as a array of Fp

crypto/src/hash/rescue_prime/mod.rs
jotabulacios: Yes, the matrix is a square matrix with `m x m` dimension. However, it's now precomputed, so this not used anymore

crypto/src/hash/rescue_prime/mod.rs
jotabulacios: Agree, I've removed the variable declarations since they weren't necessary

crypto/src/hash/rescue_prime/utils.rs
jotabulacios: Added a mention to the implementation that was followed

crypto/src/hash/rescue_prime/utils.rs
```diff
@@ -0,0 +1,75 @@
...
+        .map(|i| {
+            input.iter().enumerate().fold(Fp::zero(), |acc, (j, val)| {
+                acc + *val * omega.pow((i * j) as u64)
+            })
+        })
+        .collect()
+}
+
+pub fn intt(input: &[Fp], omega_inv: Fp) -> Vec<Fp> {
+    let inv_n = Fp::from(input.len() as u64).inv().unwrap();
```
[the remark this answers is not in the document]
jrchatruc: Let's remove the old commented version of the function here

crypto/src/hash/rescue_prime/parameters.rs
jrchatruc: &'static is correct here since the function is literally returning references to static (a.k.a global) arrays

crypto/src/hash/rescue_prime/rescue_prime_optimized.rs
```diff
@@ -0,0 +1,826 @@
...
+                get_mds_matrix(SecurityLevel::Sec128),
+                get_round_constants(SecurityLevel::Sec128),
+                get_mds_vector(SecurityLevel::Sec128),
+            ),
+            SecurityLevel::Sec160 => (
+                get_mds_matrix(SecurityLevel::Sec160),
+                get_round_constants(SecurityLevel::Sec160),
+                get_mds_vector(SecurityLevel::Sec160),
+            ),
+        };
```
ilitteri, line 72: This match is not needed anymore.
```suggestion
        let mds_matrix = get_mds_matrix(security_level);
        let mds_constants = get_round_constants(security_level);
        let mds_vector = get_mds_vector(security_level);
```

crypto/src/hash/rescue_prime/rescue_prime_optimized.rs
```diff
@@ -0,0 +1,826 @@
...
+    }
+}
+
+impl RescuePrimeOptimized {
+    /// Creates a new instance of `RescuePrimeOptimized` with corresponding Security level and the specified MDS method.
+    pub fn new(security_level: SecurityLevel, mds_method: MdsMethod) -> Result<Self, &'static str> {
+        let (m, capacity) = match security_level {
+            SecurityLevel::Sec128 => (12, 4),
+            SecurityLevel::Sec160 => (16, 6),
+        };
```
ilitteri, line 59: Let's do the same for these. `get_state_size(security_level: SecurityLevel)` and `get_capacity(security_level: SecurityLevel)`.

ilitteri (commented):
The PR is looking good! Left some comments

crypto/src/hash/rescue_prime/rescue_prime_optimized.rs
jotabulacios: You're right, thanks. Removed the match here and for size and capacity

pablodeymo (approved):

ilitteri (approved):
Great work! LGTM

PR #931, opened 2024-10-16
Add from Circom to Lambdaworks example

jotabulacios:
# Add from Circom to Lambdaworks example

## Description

Small PR adding a walkthrough tutorial  to create a circuit in Circom, generate a proof, and verify it using with Groth16 using Lambdaworks


examples/prove-verify-circom/circom_lambdaworks_tutorial.md
```diff
@@ -0,0 +1,284 @@
...
+```
+Here
+`r1cs` generates the R1CS file
+`wasm` generates the WebAssembly code
+`-p bls12381` sets the elliptic curve.
+
+That will create a `fibonacci_js` directory and a `fibonacci_r1cs `file
+
+> [!WARNING]
+> Do not skip the -p bls12381 flag as this is the only field supported by the adapter right now. If not specified the default is bn128 and the proving will not be possible
```
diegokingston, line 103: We should add support for bn254 eventually

diegokingston (approved):

ColoCarletti (approved):

PR #941, opened 2024-11-23
Fix minor grammatical errors in documentation

famouswizard:
# Fix minor grammatical errors in documentation

## Description

Two minor grammatical issues in the documentation files `docs/src/fft/benchmarks.md` and `docs/src/plonk/implementation.md`. These errors could cause confusion for readers and reduce the professionalism of the documentation.

1. In `docs/src/fft/benchmarks.md`, the sentence:
   > "Those cases which were greater than 30 seconds were marked respectively as they're too slow and weren't worth to be benchmarked."
   was corrected to:
   > "Those cases which were greater than 30 seconds were marked respectively as they're too slow and weren't worth benchmarking."
   **Explanation**: The phrase "weren't worth to be benchmarked" is grammatically incorrect. The correct phrase is "weren't worth benchmarking," which improves the readability and clarity of the sentence.

2. In `docs/src/plonk/implementation.md`, the sentence:
   > "Let's brake it down."
   was corrected to:
   > "Let's break it down."
   **Explanation**: The word "brake" was mistakenly used instead of "break." "Brake" refers to a stopping mechanism, while "break" is the correct word in this context, meaning to divide or analyze something in parts.

These changes improve the clarity and professionalism of the documentation. 

## Type of change

- [x] Optimization

## Checklist

### DIFF
--- a/docs/src/fft/benchmarks.md
+++ b/docs/src/fft/benchmarks.md
@@ -10,7 +10,7 @@ Three methods of polynomial interpolation were benchmarked, with different input
 
 these were run with criterion-rs in a MacBook Pro M1 (18.3), statistically measuring the total run time of one iteration. The field used was a 256 bit STARK-friendly prime field.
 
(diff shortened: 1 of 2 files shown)


ilitteri:
Thank you for pointing out this error, but we do not accept external contributions on typo fixes.

PR #944, opened 2024-11-29
Fix Cargo clippy 

nicole-graus:
Fix: curves, unsigned integer and hash clippy.

### DIFF
--- a/crypto/src/hash/poseidon/mod.rs
+++ b/crypto/src/hash/poseidon/mod.rs
@@ -70,9 +70,6 @@ impl<P: PermutationParameters> Poseidon for P {
         x.clone()
     }
 
-    // Note: We can't follow the advice of clippy of using div_ceil in the line 82
-    // because if we do so test_hash_many() fails.
-    #[expect(clippy::manual_div_ceil)]
     fn hash_many(inputs: &[FE<Self::F>]) -> FE<Self::F> {
         let r = P::RATE; // chunk size
         let m = P::STATE_SIZE; // state size
@@ -80,7 +77,7 @@ impl<P: PermutationParameters> Poseidon for P {
         // Pad input with 1 followed by 0's (if necessary).
         let mut values = inputs.to_owned();
         values.push(FE::from(1));
-        values.resize(((values.len() + r - 1) / r) * r, FE::zero());
+        values.resize(values.len().div_ceil(r) * r, FE::zero());
 
         assert!(values.len() % r == 0);
         let mut state: Vec<FE<Self::F>> = vec![FE::zero(); m];
--- a/crypto/src/hash/poseidon/mod.rs
+++ b/crypto/src/hash/poseidon/mod.rs
@@ -70,7 +70,9 @@ impl<P: PermutationParameters> Poseidon for P {
         x.clone()
     }
 
-    #[allow(clippy::manual_div_ceil)]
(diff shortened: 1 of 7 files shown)


crypto/src/hash/poseidon/mod.rs
```diff
@@ -70,6 +70,7 @@ impl<P: PermutationParameters> Poseidon for P {
         x.clone()
     }
 
+    #[allow(clippy::manual_div_ceil)]
```
Oppen, line 73: Why not just follow the advice? It seems it's complaining about line 80:
```rust
values.resize(((values.len() + r - 1) / r) * r, FE::zero());
```
This could be:
```rust
values.resize(values.len().div_ceil(r), FE::zero());
```

Oppen: Also, the `assert` below should probably go (the `_ceil` part implies it) or at least become `debug_assert` instead.

crypto/src/hash/sha3/mod.rs
```diff
@@ -13,6 +13,7 @@ impl Sha3Hasher {
         Self
     }
 
+    #[allow(clippy::manual_div_ceil)]
```
Oppen, line 16: Similarly, in line 19 we can simplify to:
```rust
let ell = len_in_bytes.div_ceil(b_in_bytes);
```

crypto/src/hash/poseidon/mod.rs
Oppen: Actually, it's:
```rust
values.resize(values.len().div_ceil(r) * r, FE::zero());
```

crypto/src/hash/sha3/mod.rs
```diff
@@ -16,7 +16,7 @@ impl Sha3Hasher {
     pub fn expand_message(msg: &[u8], dst: &[u8], len_in_bytes: u64) -> Result<Vec<u8>, String> {
         let b_in_bytes = Sha3_256::output_size() as u64;
 
-        let ell = (len_in_bytes + b_in_bytes - 1) / b_in_bytes;
+        let ell = (len_in_bytes + b_in_bytes - 1).div_ceil(b_in_bytes);
```
Oppen, line 19:
```suggestion
        let ell = len_in_bytes.div_ceil(b_in_bytes);
```

crypto/src/hash/poseidon/mod.rs
nicole-graus: Yes, we fixed it [here](https://github.com/lambdaclass/lambdaworks/pull/944/commits/cf35df8e506cd61bf54d872de4002b99f64ce9a3).

crypto/src/hash/sha3/mod.rs
nicole-graus: Same as before, it is [here](https://github.com/lambdaclass/lambdaworks/pull/944/commits/cf35df8e506cd61bf54d872de4002b99f64ce9a3).

crypto/src/hash/sha3/mod.rs
nicole-graus: Same as before, it is fixed [here](https://github.com/lambdaclass/lambdaworks/pull/944/commits/cf35df8e506cd61bf54d872de4002b99f64ce9a3).

Oppen (approved):

ColoCarletti (approved):

PR #945, opened 2024-11-29
Refactor STARK for small fields

ColoCarletti:
# Refactor STARK for small fields

This PR refactors the `Air` and `TransitionConstraints` traits to make the STARKs protocol compatible with small fields and their extensions.

## Type of change

- [ ] New feature
- [x] Bug fix
- [ ] Optimization

### DIFF
--- a/provers/stark/src/constraints/evaluator.rs
+++ b/provers/stark/src/constraints/evaluator.rs
@@ -3,7 +3,7 @@ use super::boundary::BoundaryConstraints;
 use crate::debug::check_boundary_polys_divisibility;
 use crate::domain::Domain;
 use crate::trace::LDETraceTable;
-use crate::traits::AIR;
+use crate::traits::{TransitionEvaluationContext, AIR};
 use crate::{frame::Frame, prover::evaluate_polynomial_on_lde_domain};
 use itertools::Itertools;
 #[cfg(not(feature = "parallel"))]
@@ -14,6 +14,7 @@ use rayon::{
     iter::IndexedParallelIterator,
     prelude::{IntoParallelIterator, ParallelIterator},
 };
+
 #[cfg(feature = "instruments")]
 use std::time::Instant;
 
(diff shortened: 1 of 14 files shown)


diegokingston (approved):

provers/stark/src/examples/bit_flags.rs
```diff
@@ -45,11 +44,22 @@ impl TransitionConstraint<StarkField, StarkField> for BitConstraint {
...
+                frame,
+                periodic_values,
+                rap_challenges,
+            }
+            | TransitionEvaluationContext::Verifier {
+                frame,
+                periodic_values,
+                rap_challenges,
+            } => (frame, periodic_values, rap_challenges),
+        };
```
entropidelic, line 61: could this pattern be made into a method of the `TransitionEvaluationContext` maybe? I know the generics may bother the compiler so if this is troublesome then nevermind, we can think of it later.

provers/stark/src/constraints/evaluator.rs
```diff
@@ -184,7 +185,11 @@ impl<A: AIR> ConstraintEvaluator<A> {
 
                 // Compute all the transition constraints at this point of the LDE domain.
                 let evaluations_transition =
-                    air.compute_transition_prover(&frame, &periodic_values, rap_challenges);
+                    air.compute_transition(&TransitionEvaluationContext::Prover {
+                        frame: &frame,
+                        periodic_values: &periodic_values,
+                        rap_challenges,
+                    });
```
entropidelic, line 192: very small nit, but it may be good to have two methods `TransitionEvaluationContext::new_prover_context(frame, periodic_values, rap_challenges)` and `TransitionEvaluationContext::new_verifier_context(frame, periodic_values, rap_challenges)`, (or maybe something with a better name) instantiate it before calling compute transition and then passing it. It makes a little easier to reason about it while reading the code :). Same comment as before, if this adds complexity by the compiler bothering about generics then leave it as is

entropidelic (approved):
lgtm! really liked the `TransitionEvaluationContext` solution

provers/stark/src/examples/bit_flags.rs
entropidelic: yeah I suspected it, not really important to do right now

PR #959, opened 2025-01-11
Fix Groth16 Circom Adapter

### LINKED ISSUE #958: bug: groth16 circom-adapter incorrect public signals
Hi, I am confident that the public signals are handled incorrectly in the [Groth16 circom adapter](https://github.com/lambdaclass/lambdaworks/blob/main/provers/groth16/circom-adapter).

First, lets check the vitalik integration test [here](https://github.com/lambdaclass/lambdaworks/blob/main/provers/groth16/circom-adapter/src/integration_tests.rs#L43). It belongs to the circuit:

```cs
template Test() {
	signal input x;
	signal output out;

	signal sym_1;
	signal y;

	sym_1 <== x * x;
	out <== (sym_1 * x) + (x + 5);
}
```

The circuit code above shows that we have 1 output (which is a public signal), 0 public inputs and 1 private inputs; these are also confirmed by the r1cs file [here](https://github.com/lambdaclass/lambdaworks/blob/main/provers/groth16/circom-adapter/test_files/vitalik_example/test.r1cs.json#L6).

As the [test asserts](https://github.com/lambdaclass/lambdaworks/blob/main/provers/groth16/circom-adapter/src/integration_tests.rs#L55) the witness returned from `circom_to_lambda` is as follows:

```ts
["1", "3", "23", "9"] // these are all hexadecimal btw
```

As per the README & the poseidon test [here](https://github.com/lambdaclass/lambdaworks/blob/main/provers/groth16/circom-adapter/src/integration_tests.rs#L23), the public signals are returned by the slice `&witness[..qap.num_of_public_inputs]`. There are two things wrong with this:

1. The public output was not considered, so `num_of_public_inputs` is only 1 due to the constant 1 added in [here](https://github.com/lambdaclass/lambdaworks/blob/main/provers/groth16/circom-adapter/src/lib.rs#L23).
2. Even if the outputs were considered, we still have to consider two slices, e.g. one for the constant `1` and other for the `23` that skips over the private inputs, in other words we care about: `["1", ..., "23", ...]` within our witness but we have no way to do this without knowing the number of non-intermediate private inputs.

I think there should be two things done:

- I believe the correct form of pub input calculation should be:

```rs
let num_of_pub_inputs =
    circom_r1cs["nPubInputs"].as_u64().unwrap() as usize +
    circom_r1cs["nOutputs"].as_u64().unwrap() as usize +
    1;
```

- The public signals should not be `&witness[..qap.num_of_public_inputs]` but instead slice in such a way that it skips over private inputs. Perhaps a QAP method can take a reference to the witness and do this, without letting the user do the slicing manually like this.

However, the QAP is edited in such a way that the proof verification actually passes with the wrong public inputs, as the passing tests show. I have not yet looked deep into what might change there, but would love to contribute if shown a hint 🙏🏻

erhant:
# Fix Groth16 Circom Adapter

## Description

1. This PR fixes #958 which is mainly due to the way public signals were handled in Circom adapter. It appears that we can keep the Circom ordering as-is, without adjusting LRO or witness signals which used to do:

- Circom: `["1", ..outputs, ...inputs, ...other_signals]`
- Lambda: `["1", ...inputs, ..outputs,  ...other_signals]`

2. We also change the way public signals are calculated, before it only cared about public inputs but now it counts for public inputs and public outputs as well.

3. `circom_to_lambda` now returns the public signals as well, to not let the user shoot themselves in the foot while computing it as it is different than expected with the addition of constant term at the start. (e.g. if public signals are `a, b` you will actually get `1, a, b`.

4. We add a `CircomR1CS` type which helps with several things:
  - we use a well-defined type instead of generic `Value` with many `as_TYPE().unwrap() as TYPE`s around the place
  - we add deserialization for the `FrElement` within the object itself
  - we use `BufReader` with `serde_json::from_reader` instead of `from_str` for performance

5. Removed some redundant `clone`s and tidied up some parts of the code.

6. Docs will be updated 🚧 

7. Changed the folder structure to be more inline with Cargo recommendations, mainly that `tests` folder has the integration tests and one test file per test scenario.

## Type of change

- [x] New feature
- [x] Bug fix
- [x] Optimization

## Checklist
- [x] Linked to Github Issue: 
- [x] Unit tests added
  - [ ] Benchmarks added/run


erhant:
I will create a new PR for the new structure, using `git mv` for moved files 🙏🏻

PR #964, opened 2025-01-24
Update README.md

MaxweLL22-22:
Original: "witnesses, and this vector of field elements is called witness alltogether."
Corrected: "witnesses, and this vector of field elements is called witness altogether."


jotabulacios:
Thank you for your contribution. However, we currently don’t accept external PRs for minor typo fixes.

PR #967, opened 2025-01-31
fix: fixed incorrect method argument type in prover.rs

brawncode:
# TITLE

## Description

In the code, there was a mismatch in the type used for the method `compute_trace_polys_main`.
The method was being called with `A::Field`, but it should use `A::FieldExtension` to match the expected type in the rest of the code. The corrected line is:

```rust
let trace_polys = trace.compute_trace_polys_main::<A::FieldExtension>();
```

## Type of change

- [ ] New feature
- [] Bug fix
- [x] Optimization

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run

### DIFF
--- a/provers/stark/src/prover.rs
+++ b/provers/stark/src/prover.rs
@@ -222,7 +222,7 @@ pub trait IsStarkProver<A: AIR> {
         A::Field: IsSubFieldOf<A::FieldExtension>,
     {
         // Interpolate columns of `trace`.
-        let trace_polys = trace.compute_trace_polys_main::<A::Field>();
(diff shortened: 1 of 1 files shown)


nicole-graus:
Thank you for wanting to contribute to Lambdaworks. The method `compute_trace_polys_main()` should use the base field, that is `A::Field`. Using the field extension would make all the operations slower, since the values of the main trace are base field elements.

nicole-graus:
Thank you for wanting to contribute to Lambdaworks. The method `compute_trace_polys_main()` should use the base field, that is `A::Field`. Using the field extension would make all the operations slower, since the values of the main trace are base field elements.

PR #970, opened 2025-02-11
Change elliptic curve new point function

nicole-graus:
# Change elliptic curve new point function

## Description

This PR changes the function `new` for every elliptic curve in Lambdaworks. In the previous version, this function creates a new curve point but doesn't check if it satisfies the curve equation. In this PR the function checks if the point is valid returning a `Result`.

### DIFF
--- a/math/benches/elliptic_curves/bn_254.rs
+++ b/math/benches/elliptic_curves/bn_254.rs
@@ -26,22 +26,17 @@ type G2 = ShortWeierstrassProjectivePoint<BN254TwistCurve>;
 #[allow(dead_code)]
 pub fn bn_254_elliptic_curve_benchmarks(c: &mut Criterion) {
     let mut rng = StdRng::seed_from_u64(42);
+    let a_val: u128 = rng.gen();
+    let b_val: u128 = rng.gen();
+    let a_g1 = BN254Curve::generator().operate_with_self(a_val).to_affine();
+    let b_g1 = BN254Curve::generator().operate_with_self(b_val).to_affine();
 
-    // let a_val: u128 = rng.gen();
-    // let b_val: u128 = rng.gen();
-    let a_val: u128 = 2;
-    let b_val: u128 = 93;
-    let p = BN254Curve::generator();
-    let q = BN254TwistCurve::generator();
-    let a_g1 = BN254Curve::generator().operate_with(&p);
-    let b_g1 = BN254Curve::generator().operate_with_self(b_val);
-
-    //let a_g2 = BN254TwistCurve::generator().operate_with_self(2u32);
-    let a_g2 = BN254TwistCurve::generator().operate_with_self(2u32);
-    assert!(a_g2.is_in_subgroup(), "a_g2 is not in the subgroup!");
-
-    let b_g2 = BN254TwistCurve::generator().operate_with_self(b_val);
-
+    let a_g2 = BN254TwistCurve::generator()
+        .operate_with_self(b_val)
+        .to_affine();
+    let b_g2 = BN254TwistCurve::generator()
+        .operate_with_self(b_val)
+        .to_affine();
     let f_12 = Fp12E::from_coefficients(&[
         "1", "2", "3", "4", "5", "6", "7", "8", "9", "10", "11", "12",
     ]);
--- a/math/benches/elliptic_curves/bn_254.rs
+++ b/math/benches/elliptic_curves/bn_254.rs
@@ -28,21 +28,17 @@ pub fn bn_254_elliptic_curve_benchmarks(c: &mut Criterion) {
     let mut rng = StdRng::seed_from_u64(42);
     let a_val: u128 = rng.gen();
     let b_val: u128 = rng.gen();
-    let a_g1 = BN254Curve::generator().operate_with_self(a_val).to_affine();
-    let b_g1 = BN254Curve::generator().operate_with_self(b_val).to_affine();
+    let a_g1 = BN254Curve::generator().operate_with_self(a_val);
+    let b_g1 = BN254Curve::generator().operate_with_self(b_val);
 
-    let a_g2 = BN254TwistCurve::generator()
-        .operate_with_self(b_val)
-        .to_affine();
-    let b_g2 = BN254TwistCurve::generator()
-        .operate_with_self(b_val)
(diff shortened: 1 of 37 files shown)


math/src/elliptic_curve/edwards/traits.rs
```diff
@@ -15,4 +17,16 @@ pub trait IsEdwards: IsEllipticCurve + Clone + Debug {
             - FieldElement::<Self::BaseField>::one()
             - Self::d() * x.pow(2_u16) * y.pow(2_u16)
     }
+
+    // Edwards equation in projective coordinates.
+    // a * x^2 * z^2 + y^2 * z^2 - z^4 = d * x^2 * y^2
```
diegokingston, line 22: This could be more efficient just by factoring out z^2. We can update that in future PR

diegokingston (approved):

math/src/elliptic_curve/edwards/curves/bandersnatch/curve.rs
```diff
@@ -15,15 +15,18 @@ impl IsEllipticCurve for BandersnatchCurve {
...
-        Self::PointRepresentation::new([
-            FieldElement::<Self::BaseField>::new_base(
-                "29C132CC2C0B34C5743711777BBE42F32B79C022AD998465E1E71866A252AE18",
-            ),
-            FieldElement::<Self::BaseField>::new_base(
-                "2A6C669EDA123E0F157D8B50BADCD586358CAD81EEE464605E3167B6CC974166",
-            ),
-            FieldElement::one(),
-        ])
+        unsafe {
```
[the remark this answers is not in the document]
Oppen: Same for other uses of `unsafe`.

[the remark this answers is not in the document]
jotabulacios: Thanks! Solved in [2e25e16](https://github.com/lambdaclass/lambdaworks/pull/970/commits/2e25e163204211d08f86f18aca2edbdc3a059895)

Oppen (commented):
Overall it looks correct, however:
1. Did we check that the use of `.unwrap_unchecked()` actually provides any advantage over regular `.unwrap()`?
2. I would suggest, regardless of which kind of `unwrap` we use, that we do something like the following:
```rust
let val = Curve::new(...);
debug_assert!(val.is_some());
val.unwrap()
```
That way we can make sure the condition holds in debug builds, without introducing any checking overhead in release builds.

math/benches/elliptic_curves/bn_254.rs
```diff
@@ -28,11 +28,15 @@ pub fn bn_254_elliptic_curve_benchmarks(c: &mut Criterion) {
...
-    let b_g2 = BN254TwistCurve::generator().operate_with_self(b_val);
+    let a_g1 = BN254Curve::generator().operate_with_self(a_val).to_affine();
+    let b_g1 = BN254Curve::generator().operate_with_self(b_val).to_affine();
+
+    let a_g2 = BN254TwistCurve::generator()
+        .operate_with_self(b_val)
+        .to_affine();
+    let b_g2 = BN254TwistCurve::generator()
+        .operate_with_self(b_val)
+        .to_affine();
```
Oppen, line 39: I believe this changes the whole meaning of the benchmark. E.g., the first two bench runs are precisely for converting to affine, which becomes trivial with the change.

math/src/elliptic_curve/edwards/curves/bandersnatch/curve.rs
```diff
@@ -12,18 +12,38 @@ impl IsEllipticCurve for BandersnatchCurve {
...
+    /// - This function should **never** be modified unless the new generator is fully verified.
     fn generator() -> Self::PointRepresentation {
-        Self::PointRepresentation::new([
+        // SAFETY:
+        // - The generator point coordinates (x, y) are taken from a well-tested,
+        //   verified implementation.
+        // - The constructor will only fail if the values are invalid, which is
+        //   impossible given that they are constants taken from a trusted source.
+        // - `unwrap_unchecked()` avoids unnecessary checks, as we guarantee
+        //   correctness based on external verification.
```
Oppen, line 35:
```suggestion
```

No longer relevant, we now use the regular `unwrap`.

math/src/elliptic_curve/edwards/curves/bandersnatch/curve.rs
```diff
@@ -12,18 +12,38 @@ impl IsEllipticCurve for BandersnatchCurve {
...
+    /// Returns the generator point of the Bandersnatch curve.
+    ///
+    /// The generator point is defined with coordinates `(x, y, 1)`, where `x` and `y`
+    /// are precomputed constants that belong to the curve.
+    ///
+    /// # Safety
+    ///
+    /// - The generator values are taken from the [Arkworks implementation](https://github.com/arkworks-rs/curves/blob/5a41d7f27a703a7ea9c48512a4148443ec6c747e/ed_on_bls12_381_bandersnatch/src/curves/mod.rs#L120)
+    ///   and have been converted to hexadecimal.
+    /// - `unwrap_unchecked()` is safe because:
```
Oppen, line 24:
```suggestion
    /// - `unwrap()` does not panic because:
```

math/src/elliptic_curve/edwards/curves/bandersnatch/curve.rs
```diff
@@ -12,18 +12,38 @@ impl IsEllipticCurve for BandersnatchCurve {
...
             FieldElement::<Self::BaseField>::new_base(
                 "29C132CC2C0B34C5743711777BBE42F32B79C022AD998465E1E71866A252AE18",
             ),
             FieldElement::<Self::BaseField>::new_base(
                 "2A6C669EDA123E0F157D8B50BADCD586358CAD81EEE464605E3167B6CC974166",
             ),
             FieldElement::one(),
-        ])
+        ]);
+        debug_assert!(point.is_ok());
```
Oppen, line 45:
```suggestion
```
No longer necessary, `.unwrap()` has the exact same effect.

Oppen (commented):
Added a few suggestions. Those apply to other places where we changed the `unwrap_unchecked`s to `unwrap`s as well.

Oppen (approved):

PR #974, opened 2025-03-06
Add extension degree const to isField trait

ColoCarletti:
This PR adds a constant to the `isField` trait to determine the degree of an extension. In cases where it is not an extension, we define it as degree 1.


diegokingston:
This PR is no longer necessary with some changes added

PR #976, opened 2025-03-11
Fix sample field element for default transcript.

ColoCarletti:
This PR fixes an issue where creating a field element from a random sample for a default transcript, used in the Fiat-Shamir protocol, did not produce elements with a uniform distribution.

In this PR, a new trait HasDefaultTranscript is introduced. It implements a function that handles the necessary logic to achieve uniform sampling, following the recommendations from: [The Definitive Guide to Modulo Bias and How to Avoid It](https://research.kudelskisecurity.com/2020/07/28/the-definitive-guide-to-modulo-bias-and-how-to-avoid-it/).

The implemented method extracts the required number of bits from a 32-byte sample to represent the field's order and constructs the field element. If a reduction is needed, the sample is rejected, and a new one is requested. The rejection frequency depends on the difference between the field's order and the nearest power of 2. In the worst case, samples are rejected 50% of the time, though for most commonly used fields, the rejection rate is much lower.

- [ ] New feature
- [x] Bug fix
- [ ] Optimization

## Checklist
- [ ] Linked to Github Issue
- [x] Unit tests added
  - [ ] Benchmarks added/run

### DIFF
--- a/crypto/src/fiat_shamir/default_transcript.rs
+++ b/crypto/src/fiat_shamir/default_transcript.rs
@@ -63,9 +63,8 @@ where
 
     fn sample_field_element(&mut self) -> FieldElement<F> {
         loop {
-            match F::get_random_field_element_from_seed(self.sample()) {
-                Some(result) => return result,
-                None => continue,
+            if let Some(result) = F::get_random_field_element_from_seed(self.sample()) {
+                return result;
             }
         }
     }
--- a/crypto/src/fiat_shamir/default_transcript.rs
+++ b/crypto/src/fiat_shamir/default_transcript.rs
@@ -1,19 +1,19 @@
 use super::is_transcript::IsTranscript;
 use core::marker::PhantomData;
 use lambdaworks_math::{
-    field::{element::FieldElement, traits::IsField},
+    field::{element::FieldElement, traits::{HasDefaultTranscript, IsField}},
     traits::ByteConversion,
 };
 use sha3::{Digest, Keccak256};
 
-pub struct DefaultTranscript<F: IsField> {
+pub struct DefaultTranscript<F: HasDefaultTranscript> {
     hasher: Keccak256,
     phantom: PhantomData<F>,
 }
 
 impl<F> DefaultTranscript<F>
 where
-    F: IsField,
+    F: HasDefaultTranscript,
     FieldElement<F>: ByteConversion,
 {
     pub fn new(data: &[u8]) -> Self {
@@ -36,7 +36,7 @@ where
 
 impl<F> Default for DefaultTranscript<F>
 where
-    F: IsField,
+    F: HasDefaultTranscript,
     FieldElement<F>: ByteConversion,
 {
     fn default() -> Self {
@@ -46,7 +46,7 @@ where
 
 impl<F> IsTranscript<F> for DefaultTranscript<F>
 where
-    F: IsField,
+    F: HasDefaultTranscript,
     FieldElement<F>: ByteConversion,
 {
     fn append_bytes(&mut self, new_bytes: &[u8]) {
@@ -62,7 +62,12 @@ where
     }
 
     fn sample_field_element(&mut self) -> FieldElement<F> {
-        FieldElement::from_bytes_be(&self.sample()).unwrap()
+        loop {
+            match F::get_random_field_element_from_seed(self.sample()) {
+                Some(result) => return result,
+                None => continue,
+            }
+        }
     }
 
     fn sample_u64(&mut self, upper_bound: u64) -> u64 {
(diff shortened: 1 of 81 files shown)


diegokingston (approved):

math/src/field/fields/montgomery_backed_prime_fields.rs
```diff
@@ -385,6 +385,36 @@ where
     }
 }
 
+impl<M, const NUM_LIMBS: usize> HasDefaultTranscript for MontgomeryBackendPrimeField<M, NUM_LIMBS>
+where
+    M: IsModulus<UnsignedInteger<NUM_LIMBS>> + Clone + Debug,
+{
+    /// # Panics
+    ///
+    /// This function will panic if NUM_LIMBS is greater than 4.
```
Oppen, line 394: Why?

math/src/field/traits.rs
```diff
@@ -282,3 +282,11 @@ pub trait IsPrimeField: IsField {
         Some((x, neg_x))
     }
 }
+
+/// This trait is necessary for sampling a random field element with a uniform distribution.
+pub trait HasDefaultTranscript: IsField {
+    /// This function should truncates the sampled bits to the quantity required to represent the order of the base field
+    /// and returns a field element if the resulting number is less than the modulus, or None if the sampled number
+    /// is greater than the modulus.
+    fn get_random_field_element_from_seed(seed: [u8; 32]) -> Option<FieldElement<Self>>;
```
cdesaintguilhem, line 291: Idea 2: rather than giving a `seed` to this function, you could give it a _seeded RNG_ object from which the implementation can draw as many bytes as needed; this would also remove the `Option<>` from the return type. This would remove the need for Idea 1.

```suggestion
    fn get_random_field_element_from_rng(rng: impl rand::Rng) -> Option<FieldElement<Self>>;
```

With this, the implementation of `get_random_field_element_from_rng()` for u64 fields could do something like
```rust
    let mut sample = [0u8; 8];
    rng.fill(&mut sample);

    mask = /* mask from u46::MAX and MODULUS.leading_zeroes() */

    while sample &= mask >= MODULUS {
        rng.fill(&mut sample)
    }
```

math/src/field/fields/montgomery_backed_prime_fields.rs
```diff
@@ -385,6 +385,36 @@ where
...
+    fn get_random_field_element_from_seed(
+        seed: [u8; 32],
+    ) -> Option<FieldElement<MontgomeryBackendPrimeField<M, NUM_LIMBS>>> {
+        let mut sample = UnsignedInteger::from_bytes_be(&seed).unwrap();
+
+        let first_non_zero_limb_index = M::MODULUS.limbs.iter().position(|&x| x != 0).unwrap_or(0);
```
cdesaintguilhem, line 400: Is it safe to return a default index of `0` here? From what I can see, the default will be returned if _all_ limbs are zero, which means that `MODULUS` itself is zero, which shouldn't happen, no? Wouldn't it be better to panic with `.expect("modulus should be non-zero")`?

math/src/field/fields/montgomery_backed_prime_fields.rs
```diff
@@ -385,6 +385,36 @@ where
...
+
+        for i in 0..first_non_zero_limb_index {
+            sample.limbs[i] = 0;
+        }
+
+        let mask = u64::MAX >> M::MODULUS.limbs[first_non_zero_limb_index].leading_zeros();
+        sample.limbs[first_non_zero_limb_index] &= mask;
+
+        if sample > M::MODULUS {
```
cdesaintguilhem, line 409: As elsewhere, `sample == MODULUS` should be rejected.

```suggestion
        if sample >= M::MODULUS {
```

crypto/src/fiat_shamir/default_transcript.rs
```diff
@@ -62,7 +62,11 @@ where
     }
 
     fn sample_field_element(&mut self) -> FieldElement<F> {
-        FieldElement::from_bytes_be(&self.sample()).unwrap()
+        loop {
+            if let Some(result) = F::get_random_field_element_from_seed(self.sample()) {
+                return result;
+            }
+        }
```
cdesaintguilhem, line 69: Following from Idea 2, this could become:
```suggestion
        let rng = <rand_chacha::ChaChaXRng as SeedableRng>::from_seed(self.sample);
        F::get_random_field_element_from_rng(rng)
```
where `X` in `ChaChaX` is 8, 12, or 20 for your favourite security level :)

math/src/field/fields/fft_friendly/quartic_babybear.rs
```diff
@@ -330,6 +330,35 @@ impl IsFFTField for Degree4BabyBearExtensionField {
     ];
 }
 
+impl HasDefaultTranscript for Degree4BabyBearExtensionField {
+    fn get_random_field_element_from_seed(seed: [u8; 32]) -> Option<FieldElement<Self>> {
+        //Babybear Prime p = 2^31 - 2^27 + 1
+        const MODULUS: u64 = 2013265921;
```
cdesaintguilhem, line 336: I'm surprised this `const MODULUS` is not defined more globally than in this function.

cdesaintguilhem (changes requested):
Looks great guys! There's a couple of places where you have a strict inequality that shouldn't be strict.

I've also left a couple of ideas from Robin and I on how do to the sampling using an RNG rather than re-sampling the transcript every time.

math/src/field/traits.rs
cdesaintguilhem: As a extra thought from Robin: the trait function signature could also take a mutable reference to the RNG so that, in the Fiat-Shamir transform, the same RNG can be re-used to generate other parts of challenges, e.g., if several field elements are required.
```suggestion
    fn get_random_field_element_from_rng(rng: &mut impl rand::Rng) -> FieldElement<Self>;
```

math/Cargo.toml
```diff
@@ -17,6 +17,7 @@ serde_json = { version = "1.0", default-features = false, features = [
 proptest = { version = "1.1.0", optional = true }
 winter-math = { package = "winter-math", version = "0.6.4", default-features = false, optional = true }
 miden-core = { package = "miden-core", version = "0.7", default-features = false, optional = true }
+rand = { version = "0.8.5", features = ["std"] }
```
pablodeymo, line 20: it could be better to use the same version as in the root Cargo.toml file:
```
rand = { workspace = true, features = ["std"] }
```

math/src/field/fields/u64_prime_field.rs
```diff
@@ -153,6 +153,18 @@ impl<const MODULUS: u64> Deserializable for FieldElement<U64PrimeField<MODULUS>>
...
 
+impl<const MODULUS: u64> HasDefaultTranscript for U64PrimeField<MODULUS> {
+    fn get_random_field_element_from_rng(rng: &mut impl rand::Rng) -> FieldElement<Self> {
+        let mut bytes = [0u8; 8];
+        rng.fill(&mut bytes);
+        let mut sample = u64::from_be_bytes(bytes);
+        let mask = u64::MAX >> MODULUS.leading_zeros();
+        sample &= mask;
+
+        U64FieldElement::new(sample)
```
cdesaintguilhem, line 164: I think you forgot the part where you need to loop until `sample < MODULUS`; at the moment this code doesn't check that the sample is inside the right range.

cdesaintguilhem (changes requested):
Looks awesome, just one loop that was forgotten I think.

math/src/field/fields/u64_prime_field.rs
diegokingston: Yes, it's missing this:

loop {
                rng.fill(&mut sample);
                let int_sample = u64::from_be_bytes(sample) & MASK;
                if int_sample < MODULUS {
                    *coeff = FieldElement::from(int_sample);
                    break;
                }
            }

math/Cargo.toml
Oppen: Do you require the `std` feature or is it just copied from the `dev-dependencies` section? `no_std` builds break because of this feature.

Oppen (approved):

PR #984, opened 2025-03-20
Add Binary Field

jotabulacios:
# Add Binary Field

## Description

This PR introduces binary fields and extends them using a tower-based approach.

It implements arithmetic in GF(2^(2^n)) by representing each field element as a polynomial over GF(2)  

We didn't implement IsField or FieldElement in this case because we wanted elements from different field extensions to coexist within the same struct, allowing us to optimize each operation to work simultaneously across all levels.

### Operations implemented:

**Addition**: Implemented as bitwise XOR.
**Multiplication**: Performed recursively using Karatsuba optimization.
**Inversion**: Achieved via recursive methods and Fermat's little theorem.
 **Exponentiation**: Implemented using the square-and-multiply algorithm.

## Type of change

- [ ] New feature
- [ ] Bug fix
- [ ] Optimization

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run


crates/math/src/field/fields/binary/field.rs
```diff
@@ -0,0 +1,611 @@
...
+    /// Note that the order of the extension field in that level is 2^num_bits.
+    #[inline]
+    pub fn num_bits(&self) -> usize {
+        1 << self.num_level()
+    }
+
+    // Equality check
+    pub fn equals(&self, other: &Self) -> bool {
+        self.value == other.value()
+    }
```
Oppen, line 93: Just in case, did you consider a custom implementation of `Eq` instead? If you did, could you add a comment explaining why you opted for a method instead?

crates/math/src/field/fields/binary/field.rs
```diff
@@ -0,0 +1,611 @@
...
+            Ordering::Greater => {
+                let mask = (1 << other.num_bits()) - 1;
+                // Lsb of a: We get the last "b.num_bits" bits of a.
+                let low = self.value() & mask;
+                // Perform the addition (XOR) on the lower part.
+                let result_low = low ^ other.value();
+                // Msb of a: We get the remaining bits.
+                let high = self.value() >> other.num_bits();
+                // We concatenate both parts.
+                let result_value = (high << other.num_bits()) | result_low;
```
Oppen, line 153: I'm a bit confused about this. Wouldn't the high bits of `other` be `0` already, making this essentially the same as `XOR` of both?

Oppen (commented):
I leave this comment as a reminder as what is missing review from me (I need a fresher brain to look at those carefully):
- mul
- inv
- tests

Everything not in that list I just reviewed and looks quite well.

crates/math/src/field/fields/binary/field.rs
nicole-graus: It was a mistake, we fixed it [here](https://github.com/lambdaclass/lambdaworks/pull/984/commits/a782a482758d07261a3408b0dc0ca3593e6013e3). Thank you!

crates/math/src/field/fields/binary/field.rs
```diff
@@ -0,0 +1,599 @@
...
+    fn from(val: u8) -> Self {
+        TowerFieldElement::new(val as u128, 3)
+    }
+}
+
+impl Default for TowerFieldElement {
+    fn default() -> Self {
+        Self::new(0, 0)
+    }
+}
```
pablodeymo, line 382: This can be easily implemented deriving `Default` in the struct

nicole-graus: Thanks! We fixed it [here](https://github.com/lambdaclass/lambdaworks/pull/984/commits/54463d7ead4d834c6de698c924c207a4679ad38c).

diegokingston (approved):

crates/math/benches/fields/binary.rs
Oppen, line 1: Not necessarily now, but the following would be desirable for benches:
1. A iai-callgrind one that runs on the CI (we already have a job set up for that);
2. Comparisons against a different implementation.

crates/math/benches/fields/binary.rs
```diff
@@ -0,0 +1,123 @@
...
+pub fn rand_element(num_level: usize) -> TowerFieldElement {
+    let mut rng = rand::thread_rng();
+    let value = rng.gen::<u128>();
+    TowerFieldElement::new(value, num_level)
+}
+
+fn binary_add_bench(c: &mut Criterion, num_levels_a: usize, num_levels_b: usize) {
+    let mut group = c.benchmark_group("Binary TowerField add");
+    let a = rand_element(num_levels_a);
+    let b = rand_element(num_levels_b);
```
Oppen, line 16: We should probably use vectors of random elements instead, so we don't accidentally pick a biased pair that performs much better or much worse than average, for example.

crates/math/src/field/fields/binary/field.rs
```diff
@@ -0,0 +1,594 @@
...
+
+                // Step 2: Get the polynomial x_{n-1} value
+                let x_value = if self.num_level() == 1 {
+                    Self::new(1, 0)
+                } else {
+                    Self::new(1 << (self.num_bits() / 4), self.num_level() - 1)
+                };
+
+                // Step 3: Compute high_product * x_{n-1}
+                let shifted_high_product = high_product.mul(x_value);
```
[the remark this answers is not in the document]
Oppen: Don't change it though. We'll see eventually if this is a bottleneck or not.

crates/math/src/field/fields/binary/field.rs
```diff
@@ -0,0 +1,594 @@
...
+            let out_hi = delta_inverse * a_hi;
+            let out_lo = delta_inverse * a_lo_next;
+
+            // Join the parts to get the final inverse
+            Ok(out_hi.join(&out_lo))
+        }
+    }
+
+    /// Calculate power.
+    pub fn pow(&self, exp: u32) -> Self {
```
Oppen, line 236: Love this, pretty and concise.

crates/math/src/field/fields/binary/field.rs
```diff
@@ -0,0 +1,594 @@
...
+    }
+}
+
+impl<'a> Mul<&'a TowerFieldElement> for &'a TowerFieldElement {
+    type Output = TowerFieldElement;
+
+    fn mul(self, other: &'a TowerFieldElement) -> TowerFieldElement {
+        <TowerFieldElement as Mul<TowerFieldElement>>::mul(*self, *other)
+    }
+}
```
Oppen, line 317: I believe you can omit the lifetimes here, as the reference never outlives the call.
If it asks for `impl<'a>`, `impl <'_>` might work instead.

crates/math/src/field/fields/binary/field.rs
```diff
@@ -0,0 +1,594 @@
...
+        assert_eq!(b * b, b);
+    }
+
+    #[test]
+    #[allow(clippy::zero_prefixed_literal)]
+    fn mul_in_level_1() {
+        let a = TowerFieldElement::new(00, 1); // 0
+        let b = TowerFieldElement::new(01, 1); // 1
+        let c = TowerFieldElement::new(10, 1); // x
+        let d = TowerFieldElement::new(11, 1); // x + 1
```
Oppen, line 425: Maybe you want `0b00` through `0b11` here. The clippy lint was actually warning correctly if that's the case.

crates/math/src/field/fields/binary/field.rs
nicole-graus: Thanks! Fixed it [here](https://github.com/lambdaclass/lambdaworks/pull/984/commits/490ea8969e6757118163c7cca900b3581bc16824).

crates/math/src/field/fields/binary/field.rs
jotabulacios: You a right, fixed in [ca2bc78](https://github.com/lambdaclass/lambdaworks/pull/984/commits/ca2bc783bc896a8633fdfb286c72158423be20eb)
Thanks!

crates/math/benches/fields/binary.rs
jotabulacios: Fixed in [28f8103](https://github.com/lambdaclass/lambdaworks/pull/984/commits/28f81033f7cb05a347ebb3c20aba8b216531c245)

pablodeymo (approved):

PR #989, opened 2025-04-01
Add RSA and Schnorr Examples

nicole-graus:
# Add RSA and Schnorr Examples

This PR adds two examples: RSA public-key cryptosystem and Schnorr signature scheme. 

You'll also find a change in the function `sample_fr_elem()` of `groth16`, because we noticed that it would be clearer to sample from entropy rather than always sampling from the same seed.

###  ⚠️ Disclaimer
These implementations are not cryptographically secure due to non-constant time operations, so they must not be used in production. They are intended to be just educational examples.


diegokingston (approved):

examples/schnorr-signature/README.md
```diff
@@ -0,0 +1,61 @@
...
+- A hash function $H$.
+
+### Signer
+- Choose a private key $k \in \mathbb{F}_p$.
+- Compute the public key $h = g^{-k}$.
+- Sample a random $ \ell \in \mathbb{F}_p$. This element can't be reused and must be sampled every time a new message 
+- Compute $r = g^\ell \in G$
+- Compute $e = H(r || M) \in \mathbb{F}_p$, where $M$ is the message.
+- Compute $s = \ell + k \cdot e \in \mathbb{F}_p$.
+- Sends $M$ with the signatur $(s, e)$.
```
pablodeymo, line 31:
```suggestion
- Sends $M$ with the signature $(s, e)$.
```

pablodeymo (approved):

PR #994, opened 2025-04-05
fix(coset): shl overflow when getting the points

0xLucqs:
For higher degree polynomials the twiddles weren't computed correctly because computing the coset size would overflow. Fixed that by switching to u32 and added tests for higher degree polynomials


diegokingston:
Hi! Thanks for the catch. In order to merge it, we need your commits to be signed (you can configure this by adding a signing key to your profile) and force pushing them.

diegokingston (approved):

pablodeymo (changes requested):

pablodeymo (approved):

PR #995, opened 2025-04-07
Pohlig Hellman Attack Example

nicole-graus:
# Pohlig Hellman Attack Example

## Description

This PR adds an implementation that demonstrates the Pohlig-Hellman algorithm for solving the discrete logarithm problem on subgroups of elliptic curves. 

In this example we chose to work with a subgroup of the BLS12-381 Elliptic Curve of order 
$$s = 96192362849 = 11 \cdot 10177 \cdot 859267.$$

This attack is significantly more efficient than attempting to solve the discrete logarithm problem using brute-force search. Using the Pohlig-Hellman attack, the logarithm can be recovered in under one second, whereas a brute-force approach could take several days. 

This highlights the importance of avoiding groups with small primes order factorization in cryptographic applications.


examples/pohlig-hellman-attack/src/pohlig_hellman.rs
```diff
@@ -0,0 +1,240 @@
...
+                // If we couldn't find the discrete logarithm in the subgroup of order `factor`, it
+                // means that there is no solution to the discrete logarithm equation proposed.
+                // That is, there is no 'x' such that h^x = q.
+                return Err(PohligHellmanError::DiscreteLogNotFound);
+            }
+        }
+
+        // We combine the equations using the Chinese Remainder Theorem.
+        let x = chinese_remainder_theorem(&equations)?;
+        Ok(x as u128)
```
pablodeymo, line 104:
```suggestion
        chinese_remainder_theorem(&equations).and_then(|x| x as u128)
```

diegokingston (approved):

examples/pohlig-hellman-attack/src/pohlig_hellman.rs
nicole-graus: Thanks, we changed it [here](https://github.com/lambdaclass/lambdaworks/pull/995/commits/62db1253ed6573a3e2b54f9f166c161e7a919a5b).

pablodeymo (approved):

PR #997, opened 2025-04-10
feat(fft): add fast mul for fft friendly fields

0xLucqs:
According to the benchmark i added this multiplication algorithm is faster for polynomials of deg >= 2 ** 6.
It's most likely not how you want the benchmark to be added but didn't really find a perfect way


crates/math/src/fft/polynomial.rs
```diff
@@ -97,6 +97,18 @@ impl<E: IsField> Polynomial<FieldElement<E>> {
         let scaled = Polynomial::interpolate_fft::<F>(fft_evals)?;
         Ok(scaled.scale(&offset.inv().unwrap()))
     }
+
+    pub fn fast_multiplication<F: IsFFTField + IsSubFieldOf<E>>(
+        &self,
+        other: &Self,
+    ) -> Result<Self, FFTError> {
+        let domain_size = self.degree() + other.degree() + 1;
```
diegokingston, line 105: this could fail if the domain size exceeds the two-addicity

crates/math/src/field/fields/mersenne31/extensions.rs
```diff
@@ -93,6 +93,12 @@ impl IsField for Degree2ExtensionField {
     }
 }
 
+impl IsFFTField for Degree2ExtensionField {
+    const TWO_ADICITY: u64 = 31;
+    const TWO_ADIC_PRIMITVE_ROOT_OF_UNITY: Self::BaseType =
```
diegokingston, line 98: where was this value taken from?

0xLucqs: from stwo. This is the generator of the unit circle
https://github.com/starkware-libs/stwo/blob/045963c3814e605e18d3edafadc79f52de8f21bb/crates/prover/src/core/circle.rs#L203-L209

crates/math/src/fft/polynomial.rs
0xLucqs: the fft function doesn't check ?

0xLucqs: indeed it doesn't seem to check
```
running 1 test
error: test failed, to rerun pass `-p lambdaworks-math --lib`

Caused by:
  process didn't exit successfully: `/Users/lucas/lambdaworks/target/release/deps/lambdaworks_math-b6eb7c1de727402e 'elliptic_curve::short_weierstrass::curves::bls12_381::default_types::tests::test_fast_multiplication' --exact --show-output` (signal: 9, SIGKILL: kill)
  ```

Comment (0xLucqs) [reply crates/math/src/fft/polynomial.rs:105]: the domain size should be checked in the fft function wdyt ? (i can add it here as well)

File: crates/math/src/field/fields/mersenne31/extensions.rs (hunk 2 above)
Comment (diegokingston) [reply crates/math/src/field/fields/mersenne31/extensions.rs:98]: Perfect, I'd add that as a comment so people can check

File: crates/math/src/fft/polynomial.rs (hunk 1 above)
Comment (diegokingston) [reply crates/math/src/fft/polynomial.rs:105]: Yes, maybe it's best if the FFT checks the condition

**Comment by 0xLucqs** [comment]
any opinion on where the benchmarks should live and what should be benchmarked ? @diegokingston

File: crates/math/src/fft/polynomial.rs [hunk 3]
```diff
@@ -27,7 +27,9 @@ impl<E: IsField> Polynomial<FieldElement<E>> {
     ) -> Result<Vec<FieldElement<E>>, FFTError> {
         let domain_size = domain_size.unwrap_or(0);
         let len = core::cmp::max(poly.coeff_len(), domain_size).next_power_of_two() * blowup_factor;
```
0xLucqs, line 29: this is super weird, when choosing a blowup factor that is not a power of 2 it fails. Don't you think this should be changed to `log_blowup_factor` and then
```suggestion
        let len = core::cmp::max(poly.coeff_len(), domain_size).next_power_of_two() * (1 << blowup_factor);
```

crates/math/src/fft/polynomial.rs
```diff
@@ -27,7 +27,9 @@ impl<E: IsField> Polynomial<FieldElement<E>> {
...
-
+        if len.trailing_zeros() as u64 > F::TWO_ADICITY {
```
0xLucqs, line 30: also this is weird, having the `TWO_ADICITY` as u64 but then everything else is either `usize` or `u32`

crates/math/src/fft/polynomial.rs
diegokingston: We could, but this would imply changing several parts in the stark protocol

0xLucqs: don't you think it would be a useful change ?

0xLucqs: Ok then maybe for another PR

pablodeymo (approved):

diegokingston (approved):

0xLucqs:
Anything missing to merge ?

crates/math/src/fft/polynomial.rs
```diff
@@ -97,6 +99,18 @@ impl<E: IsField> Polynomial<FieldElement<E>> {
...
+        other: &Self,
+    ) -> Result<Self, FFTError> {
+        let domain_size = self.degree() + other.degree() + 1;
+        let p = Polynomial::evaluate_fft::<F>(self, 1, Some(domain_size))?;
+        let q = Polynomial::evaluate_fft::<F>(other, 1, Some(domain_size))?;
+        let r = p.into_iter().zip(q).map(|(a, b)| a * b).collect::<Vec<_>>();
+
+        Polynomial::interpolate_fft::<F>(&r)
+    }
 }
```
diegokingston, line 114: This function needs to be documented with ´´´/// ´´´. In particular, it needs to explain when it makes sense to use it instead of naive multiplication, and in which cases it returns an error. Fast multiplication may look like something that should be the default, so maybe change the name to fast_fft_multiplication. Ideally, this works best with polynomials whose highest degree is equal to a power of 2 - 1.

diegokingston:
> Anything missing to merge ?

Sorry for the delay, it was a long holiday. We were looking at it and found a small comment that needs to be addressed

PR #998, opened 2025-04-14
Refactor: safely convert `usize` to `u64` in `resize_to_next_power_of_two` without unwrap

Dimitrolito:
This PR addresses and removes a previously annotated `// TODO` in the `resize_to_next_power_of_two` function.

###  What changed:
Replaced the potentially unsafe `unwrap()` on a `usize` → `u64` conversion with a safe and idiomatic cast using `as u64`, which is guaranteed to succeed on all platforms.

###  Before:
```rust
// TODO: Remove this unwrap. This may panic if the usize can't be casted into a u64.
let col_len = col.len().try_into().unwrap();
```

###  After:
```rust
let col_len = col.len() as u64;
```

## Type of change

- [x] Optimization
- [ ] New feature
- [ ] Bug fix

## Checklist
- [ ] Linked to Github Issue
- [x] Unit tests added *(not required for this low-level utility change)*

### DIFF
--- a/crates/math/src/helpers.rs
+++ b/crates/math/src/helpers.rs
@@ -14,9 +14,7 @@ pub fn resize_to_next_power_of_two<F: IsFFTField>(
     trace_colums: &mut [alloc::vec::Vec<FieldElement<F>>],
 ) {
     trace_colums.iter_mut().for_each(|col| {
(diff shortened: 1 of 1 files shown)


diegokingston:
Sorry, we don't accept this kind of PRs at lambdaworks.

PR #1001, opened 2025-04-28
Feat/fast division

0xLucqs:
# TITLE

## Description

## Type of change

- [ ] New feature
- [ ] Bug fix
- [ ] Optimization

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run

### DIFF
--- a/crates/math/src/fft/polynomial.rs
+++ b/crates/math/src/fft/polynomial.rs
@@ -140,7 +140,7 @@ impl<E: IsField> Polynomial<FieldElement<E>> {
         let d = n - m; // Degree of the quotient
         let a_rev = self.reverse(n);
         let b_rev = divisor.reverse(m);
-        let inv_b_rev = Self::invert_polynomial_mod::<F>(&b_rev, d + 1)?;
+        let inv_b_rev = b_rev.invert_polynomial_mod::<F>(d + 1)?;
         let q = a_rev
             .fast_fft_multiplication::<F>(&inv_b_rev)?
             .truncate(d + 1)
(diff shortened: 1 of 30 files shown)


crates/math/src/fft/polynomial.rs
```diff
@@ -115,6 +116,67 @@ impl<E: IsField> Polynomial<FieldElement<E>> {
...
+        {
+            return Err(FieldError::DivisionByZero.into());
+        }
+        if n < m {
+            return Ok((Self::zero(), self.clone()));
+        }
+        let d = n - m; // Degree of the quotient
+        let a_rev = self.reverse(n);
+        let b_rev = divisor.reverse(m);
+        let inv_b_rev = Self::invert_polynomial_mod::<F>(&b_rev, d + 1)?;
```
diegokingston, line 143: I don't get why this is faster than just evaluating the divisor polynomial at the FFT points, doing batch inverse and then multiplying everything out and interpolating. The method for inversion seems a bit convoluted

0xLucqs: I don't think your suggestion works when there is a remainder

diegokingston:
Sorry for the delay. It would be good to add some unit tests for some of the functions and to add a reference/document where the idea behind the algorithm was taken. The rest looks ok, though I am a bit puzzled by the number of FFTs that could be necessary to compute the division

diegokingston (approved):

pablodeymo (approved):

PR #1020, opened 2025-08-29
Add modulus range check when creating `FieldElement` from hex

gabrielbosio:
This PR serves as a proposal to add a check for the creation of a `FieldElement` from a hex string.

See https://github.com/starknet-io/types-rs/issues/81 for further details.

Note that this is one of two alternatives mentioned. The other alternative is to add this check in the `from_hex` function from `types-rs`.


pablodeymo (approved):
good catch

Oppen (approved):

ColoCarletti (approved):

PR #1024, opened 2025-09-19
Add new unchecked

jotabulacios:
# Use new_unchecked

## Description

This PR replaces the use of `new` with `new_unchecked` in cases where the points are already known to be valid.  
It also applies `new_unchecked` to the generator point in `StarkCurve`, since it is a verified valid point


Oppen:
Do we have a benchmark?

Oppen (approved):

crates/math/src/elliptic_curve/short_weierstrass/curves/bls12_381/pairing.rs
```diff
@@ -159,7 +159,8 @@ fn double_accumulate_line(
     let x1_sq_3 = three * x1.square();
     let [x1_sq_30, x1_sq_31] = x1_sq_3.value();
 
-    t.0.value = [x3, y3, z3];
+    // SAFETY: The values `x_3, y_3, z_3` are computed correctly to be on the curve.
+    t.set_unchecked([x3, y3, z3]);
```
Oppen, line 163: How is this different from assignment?

crates/math/src/elliptic_curve/short_weierstrass/curves/stark_curve.rs
```diff
@@ -17,11 +16,12 @@ impl StarkCurve {
         x_hex: &str,
         y_hex: &str,
     ) -> ShortWeierstrassProjectivePoint<Self> {
-        ShortWeierstrassProjectivePoint(ProjectivePoint::new([
+        // SAFETY: The values `x_hex, y_hex`` should be constants valid for the curve.
+        ShortWeierstrassProjectivePoint::new_unchecked([
```
Oppen: There's nothing in the method stating that. The name doesn't confer that this might silently give wrong results. I don't this is correct without also adding `_unchecked` to the method's name.

crates/math/src/elliptic_curve/short_weierstrass/point.rs
```diff
@@ -43,6 +42,17 @@ impl<E: IsShortWeierstrass> ShortWeierstrassProjectivePoint<E> {
...
+    /// checking that the point satisfies the curve equation.
+    pub const fn new_unchecked(value: [FieldElement<E::BaseField>; 3]) -> Self {
+        Self(ProjectivePoint::new(value))
+    }
+
+    /// Changes the point coordinates without checking that it satisfies the curve equation.
+    pub fn set_unchecked(&mut self, value: [FieldElement<E::BaseField>; 3]) {
+        self.0.value = value
+    }
+
```
Oppen, line 60: These should also have the following (or similar) SAFETY comment:
```rust
// SAFETY: the caller MUST be sure the arguments correspond to a point in the curve.
// Failing to do so will result in silently wrong values.
```

crates/math/src/elliptic_curve/short_weierstrass/point.rs
```diff
@@ -454,6 +458,12 @@ impl<E: IsShortWeierstrass> ShortWeierstrassJacobianPoint<E> {
         }
     }
 
+    /// Creates an elliptic curve point giving the projective [x: y: z] coordinates without
+    /// checking that the point satisfies the curve equation.
+    pub const fn new_unchecked(value: [FieldElement<E::BaseField>; 3]) -> Self {
+        Self(JacobianPoint::new(value))
+    }
```
Oppen, line 473: This also needs a SAFETY comment explaining the requirements.

crates/math/src/elliptic_curve/short_weierstrass/curves/stark_curve.rs
jotabulacios: You're right, renamed the function in [f411abd](https://github.com/lambdaclass/lambdaworks/pull/1024/commits/f411abdf8ee0ff75ace610741d7916ba78912aaa) also improved the SAFETY comment

crates/math/src/elliptic_curve/short_weierstrass/point.rs
jotabulacios: Added SAFETY comment in [f411abd](https://github.com/lambdaclass/lambdaworks/pull/1024/commits/f411abdf8ee0ff75ace610741d7916ba78912aaa)

crates/math/src/elliptic_curve/short_weierstrass/point.rs
jotabulacios: Added SAFETY comment in [f411abd](https://github.com/lambdaclass/lambdaworks/pull/1024/commits/f411abdf8ee0ff75ace610741d7916ba78912aaa)

pablodeymo (approved):

ColoCarletti (approved):

Issue #730, opened 2023-12-22
MSM Optimizations


MauroToscano:
There are a series of optimizations that can be applied to MSM that are not leveraged right now. The following can be implemented and benched, with the fastest setted to be used as the default one:

- [ ] [Faster Batch forgery Identification algorithm](https://eprint.iacr.org/2012/549.pdf). [Implementation of Gnark](https://github.com/Consensys/gnark-crypto/blob/208eeba5a82c069ec6cb51b2f82abaa6c7c870f4/ecc/bls12-381/multiexp.go#L43)
- [ ]

rahulghangas:
@MauroToscano we would like to contribute some optimization to MSM implementation based on SOA (@Strohy will be working on this). Is it okay we open up a couple of PRs?

peter941221:
This PR is benchmark-only.

- `MSM historical windows (size N)` keeps the previous unsigned window grid (`1, 2, 4, 8, 12`) for continuity.
- `MSM signed comparison (size N)` adds a separate signed-vs-unsigned grid (`2, 3, 4, 6, 8`) for a cleaner comparison.

No runtime behavior changes here, just benchmark coverage to gather better evidence before considering any MSM default-path change.

I am opening a small PR with this benchmark split now.

peter941221:
Opened PR: #1202
https://github.com/lambdaclass/lambdaworks/pull/1202

peter941221:
I opened a separate CUDA MSM follow-up in #1205.

1. Scope: this is a narrow opt-in CUDA BLS12-381 MSM improvement. It keeps `CudaMSM::compute` stable and does not change the default CPU MSM path.

2. Validation: CUDA output matched CPU `pippenger::msm` on mixed inputs, deliberate bucket collisions, negative signed digits, zero-heavy inputs, and a deterministic 257-point case.

3. Large-input signal: on my local RTX 5090-class machine, 65,536 points dropped from 225.135 ms to 74.398 ms on Windows and from 221.848 ms to 82.839 ms on WSL. At 1,048,576 points, runtime dropped from 3655.832 ms to 187.013 ms on Windows and from 2582.938 ms to 233.516 ms on WSL.

4. Small inputs are still launch/sync dominated, so I would keep this as an opt-in CUDA improvement rather than a default-path switch.

PR #1202, opened 2026-05-16
Add signed MSM benchmark comparison grid

peter941221:
## Summary
- preserve the historical unsigned MSM benchmark window grid in its own Criterion group
- add a separate signed comparison group covering sequential and parallel signed Pippenger
- keep this PR benchmark-only with no runtime MSM behavior changes

## Benchmark layout
- `MSM historical windows (size N)` keeps the prior unsigned window grid: `1, 2, 4, 8, 12`
- `MSM signed comparison (size N)` adds a signed-vs-unsigned grid: `2, 3, 4, 6, 8`

This split keeps older unsigned runs comparable while making the signed comparison easier to read in Criterion reports.

## Verification
- `cargo fmt --all`
- Bench compilation is still blocked on this Windows host by an existing `pprof v0.13.0` dependency issue in the bench graph, so no repo-local `cargo bench` result is attached here.

Refs #730

### DIFF
--- a/crates/math/benches/criterion_msm.rs
+++ b/crates/math/benches/criterion_msm.rs
@@ -40,7 +40,7 @@ pub fn historical_window_msm_benchmarks_with_size(
     assert_eq!(cs.len(), points.len());
     let msm_size = cs.len();
 
-    let mut group = c.benchmark_group(format!("MSM historical windows (size {msm_size})"));
+    let mut group = c.benchmark_group(format!("MSM benchmarks with size {msm_size}"));
 
     group.bench_function("Naive", |bench| {
         bench.iter(|| black_box(naive::msm(cs, points)));
--- a/crates/math/benches/criterion_msm.rs
+++ b/crates/math/benches/criterion_msm.rs
@@ -73,6 +73,8 @@ pub fn signed_comparison_msm_benchmarks_with_size(
     let msm_size = cs.len();
 
     // Keep signed benchmarks on a separate grid so historical unsigned data stays comparable.
+    // This benchmark group is intended to compare already-existing signed MSM paths against the
+    // unsigned ones, without changing runtime defaults in the same PR.
     let mut group = c.benchmark_group(format!("MSM signed comparison (size {msm_size})"));
 
     for &window_size in window_sizes {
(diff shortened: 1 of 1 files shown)


peter941221:
Additional context from local standalone measurements on the same `lambdaworks-math` MSM functions and seeded inputs.

These are **not** repo-local `cargo bench` results: bench compilation is still blocked on my Windows host by the existing `pprof v0.13.0` dependency issue in the bench graph. I used a small release-mode harness instead to get directional data for this benchmark-only PR.

| size | sequential unsigned | sequential signed | seq speedup | parallel unsigned | parallel signed | par speedup |
| --- | --- | --- | --- | --- | --- | --- |
| 64 | 3.677900 ms (w=4) | 3.231100 ms (w=6) | 1.138x | 1.356500 ms (w=6) | 1.268600 ms (w=4) | 1.069x |
| 256 | 10.308600 ms (w=6) | 9.384400 ms (w=6) | 1.098x | 1.723300 ms (w=6) | 1.711400 ms (w=6) | 1.007x |
| 1024 | 44.153400 ms (w=8) | 41.133100 ms (w=8) | 1.073x | 3.352100 ms (w=8) | 3.775800 ms (w=8) | 0.888x |

Current takeaway from these measurements:
- signed Pippenger looks like a modest win for sequential CPU MSM
- parallel signed is roughly flat on smaller sizes and worse at 1024 in this setup

So I see this PR as benchmark coverage only, not evidence to switch default MSM paths yet.

peter941221:
Good catch on the Criterion baseline continuity point.

I pushed a follow-up commit that keeps the existing historical group name as `MSM benchmarks with size N`, so previously saved baselines for that group remain reusable.

The new name is now only used for the additional signed comparison group.

peter941221:
I pushed a small follow-up commit that only adds rationale comments to the benchmark file.

To clarify the scope, this PR is still benchmark-only. It does not implement the Constantine-tricks path or the batch forgery identification path mentioned in issue #730, and it does not change any runtime MSM default. The goal here is narrower: compare the already-existing signed MSM paths against the current unsigned ones inside the CPU benchmark entry so we have cleaner evidence before discussing any default-path change.

I also documented why the signed comparison grid uses 2, 3, 4, 6, and 8. The signed MSM code clamps the minimum window size to 2, 3 gives one odd-window comparison point, and 4, 6, and 8 sample medium windows without expanding the benchmark matrix too aggressively.

If this benchmark-only step is useful, I still see any direct implementation of the issue checklist itself as a separate follow-up.

PR #1205, opened 2026-05-19
math/cuda: speed up large-input BLS12-381 MSM

peter941221:
This speeds up the opt-in CUDA BLS12-381 MSM path for large inputs without changing `CudaMSM::compute` or default CPU MSM selection.

Summary

1. Replace the previous CUDA bucket path with a fixed-key pipeline: count, prefix-scan, scatter, task-list accumulation, and two-stage bucket reduction.

2. Keep the public API unchanged and leave CPU MSM selection untouched.

Validation

1. CUDA output was checked against CPU `pippenger::msm`.

2. Checked cases include mixed inputs, deliberate bucket collisions, small-window collisions, negative signed digits, zero-heavy inputs, and a deterministic 257-point input.

3. All checked outputs matched CPU affine coordinates.

Benchmark

1. Windows, local RTX 5090-class run:
65,536 points: CPU 225.135 ms, CUDA 74.398 ms, 3.0x faster.
1,048,576 points: CPU 3655.832 ms, CUDA 187.013 ms, 19.5x faster.

2. WSL on the same machine:
65,536 points: CPU 221.848 ms, CUDA 82.839 ms, 2.7x faster.
1,048,576 points: CPU 2582.938 ms, CUDA 233.516 ms, 11.1x faster.

3. Every benchmark row was correctness-checked against CPU output.

Caveats

1. Small inputs are still dominated by CUDA launch and host/device synchronization overhead, so this should stay an opt-in CUDA path improvement rather than a default-path switch.

2. Repo-local Windows tests are still blocked before MSM tests by the existing unconditional `pprof v0.13.0` dev-dependency issue on this target.

3. Repo-local WSL tests are still blocked before MSM tests by an unrelated `koalabear` / `rustc 1.95.0` type-checking ICE path.

4. Because of those repo-local blockers, validation used an external harness depending only on local `lambdaworks-math` with `default-features = false` and `features = ["std", "cuda"]`.

Screenshot

CUDA MSM benchmark screenshot

Part of #730

### DIFF
--- a/crates/math/src/gpu/cuda/shaders/msm/bls12_381_msm.cu
+++ b/crates/math/src/gpu/cuda/shaders/msm/bls12_381_msm.cu
@@ -3,24 +3,90 @@
 // Compiled to .ptx by build.rs (nvcc -ptx)
 // The build system walks crates/math/src/gpu/cuda/shaders/ recursively
 // for .cu files, so this file is picked up automatically.
-//
-// WARNING: The bucket_accumulation kernel has a documented race condition
-// when multiple threads write to the same bucket without synchronization.
-// This first CUDA port carries the same limitation as the Metal MSM.
-// A correctness fix (sorting-based approach) is planned as follow-up.
 
 #include "msm.cuh"
 
 extern "C" {
 
-__global__ void bucket_accumulation_bls12_381(
+__global__ void count_bucket_entries_bls12_381(
     const int *scalars,
+    unsigned int *counts,
+    const unsigned int *config
+) {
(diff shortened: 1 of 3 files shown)


peter941221:
Addressed in f8fdd29f.

1. Removed the two unused kernel registrations from `load_ptx`.

2. Dropped the stale single-point race-condition comment in `test_cuda_msm_small`.

peter941221:
Small follow-up: this PR is still limited to the large-input CUDA MSM correctness-plus-speedup path already documented in the thread. If you want the diff cut down further, I can narrow it to the smallest mergeable slice.

